PR #18, opened 2016-05-11
Use glide and local vendoring

guilhem:
* Prevent libraries changes/disappear
* Offline building
* Usage of simple on-build dockerfile


guilhem:
commands done to create it:
`glide init` to create _glide.yaml_
`glide install --strip-vcs --strip-vendor` to create vendor dir

tanji:
Hi Guilhem,

thanks, that comes at the right time since I was reading on vendoring this morning and I was not sure which tool to pick. Glide looks good.

Could you just remove the Dockerfile from the PR?
The reason why I do not use onbuild is very simple: it uses the debian jessie base image, and the resulting image is 800M, vs 300M for the alpine image.

Thanks in advance!

guilhem:
@tanji I understand this concerned, but I think this Dockerfile is only to test and "quick and dirty" test it.

300M is far too big from my point of view, I worked on a 2 phases build based on [habitus](http://www.habitus.io/) to have a first phase with build and another only for run.

But I also encounter another problem... `mysqldump` isn't included in docker image and `provision` fail without it.

tanji:
Yes, but I don't want to overwrite my file either. Maybe renaming it to Dockerfile.onbuild or something like that would work for you?

I can include mysqldump in the build, but the idea around `provision` is that it is just a proof of concept at the moment, and it should be replaced by a pure Go implementation.

guilhem:
oh nice to hear it (about pure go implementation).
@tanji let's have a look to my latest commit, if it's not good for you I will rollback your dockerfile

For information, resulting image is 49.2 MB

tanji:
Excellent. Thank you very much!

guilhem:
@tanji don't forget to update "KEY" in `.travis.yml`. you can generate it with `travis` CLI. ;)

PR #38, opened 2016-06-22
Update Dockerfile and docker-compose

hugochinchilla:
The docker-compose file was failing to start the test cluster, this fixes it.


Dockerfile
```diff
@@ -9,4 +9,5 @@ RUN rm -rf /go/src/replication-manager
 RUN mkdir /etc/replication-manager
 COPY config.toml /etc/replication-manager/config.toml
 WORKDIR /go/bin
-CMD ["replication-manager", "monitor", "--daemon"] 
+ENTRYPOINT ["replication-manager"]
+CMD ["monitor", "--daemon"]
```
tanji: OK, that's a much better solution indeed :+1:

PR #40, opened 2016-06-22
bind agent in all interfaces by default

hugochinchilla:
The monitor instance in the docker-compose file cannot check the agents because they're only listening on localhost, this allows them to be reached from the outside.


tanji:
Thanks for the PR. FYI, the Agent doesn't do anything for now. It's just a POC and could be removed in the next future, if we judge that it's useless

PR #91, opened 2016-10-01
Disable interactive mode when in daemon mode

markus456:
Implicitly disabling interactive mode in daemon mode makes sense as there's no way to manually initiate the failover without the graphical interface.


markus456:
Disabling it always isn't such a good idea if you want to just test out how the replication manager works. A better option is to document that `--daemon` requires `--interactive=false` for automated failover.

PR #137, opened 2017-02-27
Update build instructions

markus456:
Updated build instructions to use the `glide` package manager.

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -1,8 +1,11 @@
 ## Building from source
 
 * Download a Go release (Go 1.6 minimum): https://golang.org/dl/ or use your system's release if >= 1.6
-* Create a build directory: `mkdir -p ~/go`
-* Point GOPATH environment variable to this directory: `export GOPATH=~/go`
-* Clone the source: `go get github.com/tanji/replication-manager`
-* Compile and install: `go install github.com/tanji/replication-manager`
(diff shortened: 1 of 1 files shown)


tanji:
Hi Markus,

you shouldn't have to install glide and install dependencies to build replication-manager. For convenience, all the deps are vendored and should reside in the vendor directory.
If it doesn't build please let me know why. It means that vendoring is not up to date.

Thanks
Guillaume

markus456:
On CentOS 7 the build seems to fail due to an old golang version:
```
[vagrant@galera-003 ~]$ go get github.com/tanji/replication-manager
package context: unrecognized import path "context" (import path does not begin with hostname)
[vagrant@galera-003 ~]$ go install github.com/tanji/replication-manager
go/src/github.com/tanji/replication-manager/vendor/github.com/coreos/etcd/lease/leasehttp/http.go:19:2: cannot find package "context" in any of:
	/home/vagrant/go/src/github.com/tanji/replication-manager/vendor/context (vendor tree)
	/usr/lib/golang/src/context (from $GOROOT)
	/home/vagrant/go/src/context (from $GOPATH)
[vagrant@galera-003 ~]$ go version
go version go1.6.3 linux/amd64
```

After installing latest go 1.8 it seems to work. I think this pull request can be closed as it does seem to work if the required go version is updated.

tanji:
It should work with with 1.6, all versions starting from 1.6 support vendoring.

PR #176, opened 2017-07-27
Update README.md

MariaDBJones:
MRM user need PROCESS grant to access list of internal processes


tanji:
Patch doesn't merge, I will update manually, thanks

PR #189, opened 2017-11-23
Link system V init script with systemd

koleo:
Actually you can stop & start replication-manager on Debian8+ using either sysV init scripts or systemd :

```
# Using sysV
/etc/init.d/replication-manager start

# Using systemd
systemctl start replication-manager.service
```

But, If you start mrm with systemd and try to stop it with sysV init script, it fails:

```
root@myhost:~ # /etc/init.d/replication-manager stop
/etc/init.d/replication-manager: 57: kill: No such process
```
One way to solve this problem is to map init.d script with systemd. This can be achieved simply by adding /lib/lsb/init-functions to init.d script.

With this fix, example above should now display:
```
root@myhost:~ # /etc/init.d/replication-manager stop
[ ok ] Stopping replication-manager (via systemctl): replication-manager.service.
```

### DIFF
--- a/service/replication-manager-arb-basedir.init.deb7
+++ b/service/replication-manager-arb-basedir.init.deb7
@@ -34,6 +34,7 @@ chown -R $MRMUSER:$MRMGROUP $MRMDATADIR
 CMD=$MRMBASEDIR/replication-manager-arb
 PIDFILE=$MRMDATADIR/mrm.pid
 
+. /lib/lsb/init-functions
 
 start()
 {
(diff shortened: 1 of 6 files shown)


tanji:
Hi Nicolas, great catch! Would this work on CentOS as well?

tanji:
Actually I noticed your patch is for the debian files, so I'm just gonna go ahead and merge it, and I will look up how to do the same for CentOS 7 later. Thanks!

koleo:
This fix is for Debian indeed. I have not tested mrm on Centos so I do not know if there is similar problem to fix.

Thank you for your prompt response!

PR #276, opened 2019-03-20
Restrict PFS query to non-null digests

0x6d617474:
Followup to #275, this PR modifies the query to the performance_schema to prevent null values from appearing in the results set. 

While #275 fixed the leaking connection in the event the row scanner failed, here we sanitize the data passed to the scanner to prevent the scanner from failing in the first place. This also fixes spamming "Could not get PFS queries" in the log file if there is a null value in the performance_schema.


tanji:
Thanks for the PR!

PR #295, opened 2020-03-02
fix goroutine leak when connection error

vegazmaoning:
when connection to mysql, if error is not nil, conn is not close, cause goroutine leak, because in database/sql, lauch two goroutine（connectionOpener, connectionReseter）, so goroutine unrenewable  cause memory leak.
I don't know whether agree with


vegazmaoning:
mysql last pr had problem, defer should defined before return that err handle block

svaroqui:
Thank for this contribution !

PR #296, opened 2020-03-06
fix goroutine leak

vegazmaoning:
Similar last PR, loss close connection when connection error, cause goroutine leak, also can slove it by update sqlx version.

### DIFF
--- a/cluster/srv.go
+++ b/cluster/srv.go
@@ -370,10 +370,10 @@ func (server *ServerMonitor) Refresh() error {
 		return errors.New("Connection is closed, server unreachable")
 	}
 	conn, err := sqlx.Connect("mysql", server.DSN)
+	defer conn.Close()
 	if err != nil {
 		return err
 	}
-	defer conn.Close()
 	if server.ClusterGroup.conf.MxsBinlogOn {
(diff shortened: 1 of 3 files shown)


svaroqui:
But if you close the global connection some other routing  using the global connection will failed later on no, can you reproduce memory leak with 2.1 where  sqlx was updated ?

vegazmaoning:
The global connection only close when error is not nil, so may be there is no problem.
The 2.1 that with update sqlx, will not has memory leak, because close connection when  Ping error  in sqlx.Connect 
https://github.com/signal18/replication-manager/blob/2.1.0-dev/vendor/github.com/jmoiron/sqlx/sqlx.go#L627.
But I didn't used 2.1, because i see it`s deploying, i think will not has memory leak in theory

svaroqui:
2.1 pro is provisioning but you have .
replication-manager-pro monitor --help | grep orchest
   --prov-orchestrator string               onpremise|opensvc|kube|slapos|local (default "opensvc")  

2.1 osc stay on premise per default 
./replication-manager-osc monitor --help | grep orchest
      --prov-orchestrator string                             onpremise|opensvc|kube|slapos|local (default "onpremise")

PR #327, opened 2021-02-18
Failover multisource

### LINKED ISSUE #335: Inter cluster via Master to Master multi source failover
L1->R1   Cluster C1 replication source S1 domain D1 
 !--> L2->R2   Cluster C2 replication source S2  domain D2

On topology detection :
For all cluster nodes auto discovering if an other node exists with replication named source pointing to it 
Add the extra replicas node to the parent cluster 
Mark those extra nodes as ignored for failover and proxies traffic 

This assume failover of parent cluster Leader will also point the child cluster leader to the elected leader 
For child cluster leader failover freezing write to old leader is extended to stop all other named replication sources
The extra sources are dropped and recreated on the new leader taking the same parameters as the original replication source

tanji:


cluster/cluster_fail.go
```diff
@@ -655,6 +653,90 @@ func (cluster *Cluster) MasterFailover(fail bool) bool {
...
+			if parentCluster != nil {
+				myparentrplpassword = parentCluster.rplPass
+			} else {
+				cluster.LogPrintf(LvlErr, "Unable to found a monitored cluster for replication source %s ", rep.ConnectionName.String)
+				cluster.LogPrintf(LvlErr, "Moving source %s with empty password to preserve replication stream on new master", rep.ConnectionName.String)
+			}
+			// need a way to found parent replication password
+			if strings.ToUpper(rep.UsingGtid.String) == "NO" {
+
+				logs, err := dbhelper.ChangeMaster(NewMaster.Conn, dbhelper.ChangeMasterOpt{
```
terwey: I would recommend creating a `dbhelper.ChangeMasterOpt{}` above the conditionals that share all the properties and then simply refer to that one and add the required parameters that are missing and/or different.

There's a LOT of duplication.

cluster/cluster_fail.go
```diff
@@ -178,12 +171,12 @@ func (cluster *Cluster) MasterFailover(fail bool) bool {
 		dbhelper.MasterWaitGTID(cluster.master.Conn, cluster.oldMaster.GTIDBinlogPos.Sprint(), 30)
 	} else {*/
 	// Failover
-	cluster.LogPrintf(LvlInfo, "Waiting for candidate master to apply relay log")
+	cluster.LogPrintf(LvlInfo, "Waiting for candidate master %s to apply relay log", cluster.master.URL)
 	err = cluster.master.ReadAllRelayLogs()
 	if err != nil {
-		cluster.LogPrintf(LvlErr, "Error while reading relay logs on candidate: %s", err)
+		cluster.LogPrintf(LvlErr, "Error while reading relay logs on candidate %s: %s", cluster.master, err)
```
terwey: Shouldn't this be `cluster.master.Name` or similar? currently it'll print the `ServerMonitor{}` to a string directly.

svaroqui: Good catch  cluster.mater.URL is the one to use

PR #328, opened 2021-02-23
Add cluster.SetSugarState

terwey:
Switch cluster.SetState to cluster.SetSugarState which reduces code complexity

Note: some States were before marked as WARNING while the Key was set to ERR


svaroqui:
This is critical to keep the level because of this func 
(SM *StateMachine) IsFailable() bool {
	SM.Lock()
	for _, value := range *SM.OldState {
		if value.ErrType == "ERROR" {
			SM.Unlock()
			return false
		}
	}
	SM.discovered = true
	SM.Unlock()
	return true
}

The idea behind this is that there are errors for a server that do not really lead to errors at the cluster layer 

A server can not be elected as a slave is a warning while it's an error for the server it not an issue for the cluster if an other server can be elected, so any  lvlerr   cancel the failover or the switchover cluster capability

PR #339, opened 2021-04-02
[WIP] Config store preview

### LINKED ISSUE #332: External key value config store
Enable saving and restoring per cluster config variables from various datastore. 

At minimum to implement a first datastore using the replication-manager-pro arbitrator instance that can save data in embedded sqllite or external MariaBD RDBMS

In that first step  to offer to replication-manager-pro arbitrator instead of different binary , should save some precious compile time and different packages installation   ( osc release should complain for support license issue)  

The next datastore nice to have is an etcd cluster made out of 3 different replication-manager-pro monitor instances 

We should unify all this with already existing code (monitoring-save-config)
- to save and restore per cluster config variables changes  inside a toml file store inside the datadir of the cluster
- to  restore per cluster config variables from env variables  

Keep in mind that the default section and command line parameters should be seen as a way to enforce some variables settings for all clusters and should be proceed and merge at the last step 

Today it does not exist any code facility to tell if a variable was change from it's  default and it would be nice to have in same PR

terwey:


svaroqui (approved):

svaroqui:
Re, we use newer version of protobuf from embedding the k8s client api can you please  pull from develop and merge keeping everything in higher release .

PR #458, opened 2022-12-22
Itops 122/feature/fix dockerfile

AlexandrClick:


svaroqui:
v.RegisterAlias("api-user", "api-credentials") why do yo thing this is an error alias is supposed to enable old syntax the new syntax api-credentials as it's a user:password type 
--api-credentials string                               Rest API user list user:password,.. (default "admin:repman")

svaroqui:
If you change the Processlist struct it looks like you need to make sure that the SQL function is change according 

For MariaDB what you try to introduce is already reported as Time for processlist 
 
Time_ms as Time,

func GetProcesslistTable(db *sqlx.DB, version *MySQLVersion) ([]Processlist, string, error) {
	pl := []Processlist{}
	var err error
	stmt := ""
	if version.IsMariaDB() {
		//MariaDB
		stmt = "SELECT Id, User, Host, `Db` AS `db`, Command, Time_ms as Time, State, SUBSTRING(COALESCE(INFO_BINARY,''),1,1000) as Info, CASE WHEN Max_Stage < 2 THEN Progress ELSE (Stage-1)/Max_Stage*100+Progress/Max_Stage END AS Progress FROM INFORMATION_SCHEMA.PROCESSLIST WHERE command='query' ORDER BY TIME_MS DESC LIMIT 50"

PR #514, opened 2023-10-22
chore: unnecessary use of fmt.Sprintf

testwill:


svaroqui:
Thanks will for fixing this can you agree to give the copyright to signal18 Cloud under Free BSD code  ?

testwill:

PR #515, opened 2023-11-01
Update MS Teams Notification to Newer Package

ahfa92:
Updating ms teams notification to active fork from dasrick to atc0005. 

Adding configuration for MS Teams:
- Adding proxy url (for isolated environment which only allow http traffic via proxy e.g. squid.
- Adding comma separated list configuration for capturing state based on code prefix ERR|WARN|INFO (default is none)


ahfa92:

PR #516, opened 2023-11-24
add secondary handling for backup, mentioning backup type explicitly

ahfa92:
add secondary handling for backup if backup was from master, mentioning backup type explicitly (mariabackup or xtrabackup, compressed or not)


ahfa92:

PR #521, opened 2023-11-28
Delay average statistic

ahfa92:
Adding delay average statistics for failover weighting. Already got the statistic, but still waiting for the weighting calculation. Kindly suggest the weighting algorithm.
values:
delay: average delay
delayCount: counter of lag more than 0 seconds occured
slaveErrCount: counter of slave error
counter: total counter

if we want to see percentage of delay count we can do delayCount / counter and so does with slave err percentage slaveErrCount
based on the logic we should pick the lowest value available.

### DIFF
--- a/cluster/srv.go
+++ b/cluster/srv.go
@@ -39,19 +39,6 @@ import (
 	"github.com/signal18/replication-manager/utils/state"
 )
 
-// Logging Replication Delay Stat
-type DelayStat struct {
-	DelayAvg      float64 `json:"delay"`         // Average Seconds of Delay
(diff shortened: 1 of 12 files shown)


cluster/srv.go
svaroqui, line 1: Could we organize code to put most functions in cluster_stat.go file !

svaroqui (changes requested):

cluster/srv.go
ahfa92: Sure

ahfa92:
Hi @svaroqui, I already updated the code. Kindly check.

PR #584, opened 2024-05-13
Add mode of getting binlog timestamps using mysqlbinlog and go-mysql/replication

### LINKED ISSUE #579: Checking oldest binlog timestamp returned Err: Failed to parse binary log datetime string.
Replication manager write error when datetime string from mysqlbinlog not found.

ERROR   Failed to parse binary log datetime string. . Str: . Host: 172.24.47.186:3306 - mysql-bin.000010

the reason is mysqlbinlog only parse until position 512 in order to prevent parsing many lines:
"--stop-position=512"

caffeinated92:
This PR should fix #579 

## Contributor Agreement

### DIFF
--- a/cluster/srv_binlog.go
+++ b/cluster/srv_binlog.go
@@ -14,6 +14,7 @@ import (
 	"errors"
 	"fmt"
 	"os/exec"
+	"regexp"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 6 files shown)


svaroqui (commented):
Hello  both methods can be needed for example slapos will always prefer using mysqlbinlog , i would rather like a parameter that list all possible options to extract binlog and default to go-mysql as it's the only  embedded one. Also it is prefered to have choice as go mysql being implementation of the binlog protocol it could potentialy lag behind protocol changes and we wan't a way out using mysql or mariadb mysqlbinlog

caffeinated92:
I will add the function and check for the preferences

caffeinated92:
At first mysqlbinlog has different datetime after parse. But after changing to time.Local, the value is already same with go-mysql (which is the correct value)

PR #599, opened 2024-05-22
replacing credentials with asterisk

### LINKED ISSUE #573: Database credentials are leaked in API and web interface
I am currently studying the API and web interface of Replication Manager to see how I can use it to discover all the clusters that are monitored by Replication Manager, and to discover the current topology and state per cluster. I want to automatically add items to our monitoring system Zabbix for the discovered clusters and their topologies.

While looking into the API and web interface I learned that the web interface uses the API, which is great, but I also noticed that in all API endpoints that I tried, the database credentials of the `root` user and/or the `replication` user are present in the responses. I think this is a major security concern and I currently do not see why this would ever be needed. For me this would be a reason to not use the web interface or the API. I know that the API can be protected with user credentials and can run on HTTPS, but I really think the credentials should never leave the Replication Manager server.

Maybe there is a setting controlling this behavior? If not, do you agree the credentials should never be exposed like this?

caffeinated92:
## Contributor Agreement

### DIFF
--- a/cluster/srv.go
+++ b/cluster/srv.go
@@ -48,7 +48,7 @@ type ServerMonitor struct {
 	SourceClusterName           string                       `json:"sourceClusterName"` //Used to idenfied server added from other clusters linked with multi source
 	Conn                        *sqlx.DB                     `json:"-"`
 	User                        string                       `json:"user"`
-	Pass                        string                       `json:"-"`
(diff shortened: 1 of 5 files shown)


svaroqui:
Please see the master , servers and all other api call that may require similar change

PR #614, opened 2024-05-27
User management

### LINKED ISSUE #568: Warning on api password need change for no default
Using default password for api please change it

caffeinated92:
Prompt warning default user still have default password. #568


svaroqui:
APIUsers                  map[string]APIUser          `json:"apiUsers" 
this variable can be use to get a direct hash access to the admin user no need to re parse in a loop no ?

caffeinated92:
Let me check

PR #637, opened 2024-06-11
jwt-auth-variable

### LINKED ISSUE #635: JWT Token expiration should be expose
claims["exp"] = time.Now().Add(time.Hour * 48).Unix() is hardcoded in  login making sure we can config this via variables

caffeinated92:
Add JWT Expire variable to config. This will close #635


svaroqui:
auth-jwt-expire please rename to api-token-timeout

PR #638, opened 2024-06-11
Fix unauthorized login

caffeinated92:
prevent redirecting user with valid name but wrong password to empty dashboard. this will reply with wrong password instead


svaroqui:
I'm not sure did you test this with OIDC gitlab login

PR #661, opened 2024-06-25
Migrate from deprecated ioutils

caffeinated92:
Migrate from ioutils package to io and os package.


caffeinated92:
Result fr/testFailoverNoRplChecksNoSemiSync -> PASS
Result fr/testSlaReplAllSlavesDelayNoSemiSync -> PASS

PR #704, opened 2024-07-09
Add new states for MariaDB Jira Blocker and Critical issues

caffeinated92:
Caution!
This will make the alert and prevent failover/switchover for JIRA blockers:
1. https://jira.mariadb.org/browse/MDEV-28310

For better state checking, state from server will always have serverURL, and it will be added to list for same state. (previously it was ignored)

serverURL can be multiple : A:3306,B:3306,C:3306


caffeinated92:
This is the sample using mariadb 10.11.7 
image

caffeinated92:
image

caffeinated92:
The MariaDB version has bug in service for this version, source: https://jira.mariadb.org/browse/MDEV-31779  channel=StdOut cluster=fr code=MD031779 status=OPENED type=state

svaroqui:
One can see that https://jira.mariadb.org/browse/MDEV-31779  is about having sequences in schema 
and the other critical one is having on duplicate key update  

So my proposal was to have mdev-stopper.json  where we {issue:"MDEV-31779", tags "sequences"} blabla blocker should block failover only if both version and tags match  
In our case we would never block until we are able to detect workload cluster tags sequences  , onduplicatekey

caffeinated92:
Yes, currently MDEV list is only giving alert, only MDEV-28310 is blocker.
Need to check and adding configurator tags based on situation

caffeinated92:
Yes, currently MDEV list is only giving alert, only MDEV-28310 is blocker.
```
blockers := []string{
		"MDEV-28310",
	}

	for _, mdev := range blockers {
		if sl.MDevIssues.HasMdevBug(mdev) {
			return false
		}
	}

	return true
```
Need to check and adding configurator tags based on situation

caffeinated92:
On hold while focusing for finishing backups

Issue #752, opened 2024-07-22
When the slave sets maintenance, there is a problem with the status change of proxysql.


peterpan104:
This is the state when repman parameters, repman and proxysql are first start:
```
## repman parameter
proxy-servers-read-on-master = false
proxy-servers-read-on-master-no-slave = true
```
![image](https://github.com/user-attachments/assets/50cfa058-1381-4935-ac4b-262b6d4f4cae)
<img width="1320" alt="image" src="https://github.com/user-attachments/assets/bb61db46-5a3f-46b9-b0f3-3f568dd86a54">
---
I set 172.24.44.182 as maintenance,
the status of repman and proxysql is as follows:

![image](https://github.com/user-attachments/assets/e3fb1bbd-c847-461f-8cea-e3500e1cf3bd)
<img width="1384" alt="image" src="https://github.com/user-attachments/assets/cabd21a8-2089-46de-be7a-695a49c7bedf">
---
Then set 172.24.47.188 as maintenance,
the status of repman and proxysql is as follows:

![image](https://github.com/user-attachments/assets/714200a3-6e2f-4bcc-9000-2db1ce123fa3)
<img width="1377" alt="image" src="https://github.com/user-attachments/assets/10d0cb45-f38a-4aed-9c64-fb79cfa7b991">
---
Now set 172.24.47.188 as online,
the status of repman and proxysql is as follows:

![image](https://github.com/user-attachments/assets/d16dda29-6a9f-4936-b55b-a40b5029c020)
<img width="1382" alt="image" src="https://github.com/user-attachments/assets/ce236429-0215-4ed9-b0f3-b34c5f3301c0">
---
then set 172.24.43.182 as online,
the status of repman and proxysql is as follows:

![image](https://github.com/user-attachments/assets/27435db9-3c88-415e-9301-555c5b7fff4e)
<img width="1284" alt="image" src="https://github.com/user-attachments/assets/321b42b1-d33d-4089-8018-515b4c461e4e">
---
**proxy-servers-read-on-master-no-slave seems to have no effect, but proxysql status is normal**
---
Now I trigger the switchover,
the status of repman and proxysql is as follows:

![image](https://github.com/user-attachments/assets/cbb99f5e-1281-419d-b4b8-1586168278ab)
<img width="1288" alt="image" src="https://github.com/user-attachments/assets/b5761615-aec3-434c-9528-580feec8fba2">
---
I set 172.24.43.182 as maintenance,
[...]

PR #792, opened 2024-08-07
When readonmasternoslave tag is set and bootstrap is ON,  make proxysql master as reader if no valid reader

caffeinated92:
This will add alert if bootstrap is OFF, and will be working if bootstrap is ON.


caffeinated92:
This is for #752

Issue #819, opened 2024-08-28
Physical backup does not print same general log as logical backup


svaroqui:
2024/08/28 09:36:07 | INFO | [general] [SUCCESS] Finish logical backup mysqldump for: db1.bench.svc.cloud18:3306
2024/08/28 09:34:01 | INFO | [general] Request logical backup mysqldump for: db1.bench.svc.cloud18:3306

Would be noce to do the same with physical

PR #821, opened 2024-09-02
write final state of task in general log

caffeinated92:
This will write final state of tasks in general log. This will also cover #819


svaroqui:
Could be nice to double the message in job and in general log

svaroqui:
While behing in that code sending an alert with the backup log could be nice as well when it failed regarding backup task

Issue #834, opened 2024-09-05
proxy config generation api failed 


svaroqui:
After upgrading to last release , we are no more able to provision clusters caused by init container  not able to fetch config.tag.gz

From the init container or form the web after login 

```
 wget  --no-check-certificate -q -O- https://repman.datais.svc.datais:10005/static/configurator/opensvc/bootstrap |sh -x
+ '[' -z admin ]
+ '[' -z Repm@n4DataIS ]
+ '[' -z https://repman.datais.svc.datais:10005 ]
+ '[' -z db-prod-fr ]
+ '[' -z proxysql.db-prod-fr.svc.datais ]
+ '[' -z 6032 ]
+ SET -x
sh: SET: not found
+ GET='wget -q --no-check-certificate -O- --header Content-Type:application/json'
+ AUTH_DATA='{"username": "admin", "password": "Repm@n4DataIS"}'
+ wget -q --no-check-certificate -O- --header Content-Type:application/json --post-data '{"username": "admin", "password": "Repm@n4DataIS"}' --header Accept:text/html https://repman.datais.svc.datais:10005/api/login
[...]

PR #835, opened 2024-09-05
Fix failed provision in opensvc for proxies

caffeinated92:
This PR will fix failed provision in opensvc for proxies. Related to #834


caffeinated92:
@svaroqui  with this, I don't have problem in provisioning proxysql in s18-dev-1. Is there anything else that I should do?

Issue #842, opened 2024-09-10
SSL issue using mariadb client >= 11.3 when Database nodes <11.3 


svaroqui:
All MariaDB clients >= 11.3 use SSL true flag making mysqldump mysqlbinlog or mysql failed when used on a lower server release that do not enable SSL. This new zero SSL config of MariaDB should  be supported in mixed version  

!cluster.HaveDBTLSCert && !server.HasSSL() && server.IsMariaDB() && server.DBVersion.Lower("11.3") && binlogver.IsMariaDB() && binlogver.DistVersion.GreaterEqual("11.3") {. could we replace this test with server.GetSSLClientParam(tool string).

this would enable for example to inject the correct string if we push our own generated certificates to server config and chose based on tool version as well putting all that SSL logic inside one single place

Return empty string when we have detected that no parameter need to be added

Please add comment in the code

server. HaveDBTLSCert() is on when we replication-manager generate cluster cerrificate that we use the ssl tag in configurator in this case the client ssl string need to have param about the client certificate found in replication-manager cluster datadir
Link to MaraDB sergei blog about zero ssl for mariaDB >=11.3 https://mariadb.org/mission-impossible-zero-configuration-ssl/

caffeinated92:
We should probably have different solution of managing ssl tag in config

Before MariaDB 10.4 , mysql 8.0(flush privileges;reload certificates) the lack of https://mariadb.com/kb/en/flush/#flush-ssl. force us to check expired date of certificate and add the cluster trigger force restart with certificate regeneration before the rolling restart
Before MariaDB 11.3 we can have a push config without restart and run flush ssl

caffeinated92:
Using opensvc and K8S we should use secrets to store ssl certificates and generate service config that expose them via shm to files , this enable to refresh the certificates inside container without restarting the service but by just replacing the secret and then use flush ssl or flush privileges

caffeinated92:
Add doc Secure cluster in transit explaining all this and may be the janitor proxies

caffeinated92:
A new task for DBJobs Get server certificates
For replication-manager to work using SSL we have client certificates config variables .

--db-servers-tls-client-cert string                    Database TLS client certificate
--db-servers-tls-client-key string                     Database TLS client key

Insure that we use those inside configurator instead of nothing and preserved the one generated, if those client certificates a given to us via config it means it exists some server certificates that we could fetch as well for configurator to integrate and adapt to current database setting

Document in what context those parameter are used
[...]

Issue #823, opened 2024-09-02
Check presence &  version of local mysql, mysqlbinlog, mysqldump


svaroqui:
Add warning alert when rejoin is know to failed because lower version or not found of tools needed , flashback true , direct dump restore

PR #843, opened 2024-09-11
No SSL compatibility for MariaDB 11.3+ client and also add check before using tools in backup and reseed

caffeinated92:
SSL compatibility for MariaDB 11.3+ client. Related to #842 and #823


svaroqui (commented):
!cluster.HaveDBTLSCert && !server.HasSSL() && server.IsMariaDB() && server.DBVersion.Lower("11.3") && binlogver.IsMariaDB() && binlogver.DistVersion.GreaterEqual("11.3") {.    could we replace this test with server.GetSSLClientParam(tool string).  

this would enable for example to inject the correct string if we push our own generated certificates to server config and chose based on tool version as well putting all that SSL logic inside one single place 

Return empty string when we have detected that no parameter need to be added 

Please add comment in the code  
- server. HaveDBTLSCert()  is on when we replication-manager generate cluster cerrificate that we use the ssl tag in configurator in this case the client ssl string need to have param about the client certificate found in replication-manager cluster datadir 
- Link to MaraDB sergei blog about zero ssl for mariaDB >=11.3  https://mariadb.org/mission-impossible-zero-configuration-ssl/

svaroqui:
We should probably have different solution of managing ssl tag in config 
- Before MariaDB 10.4 , mysql 8.0(flush privileges;reload certificates) the lack of https://mariadb.com/kb/en/flush/#flush-ssl. force us to check expired date of certificate and add the cluster trigger force restart with certificate regeneration before the rolling restart 
- Before MariaDB 11.3 we can have a push config without restart and run flush ssl  
- >= MariaDB 11.3 do nothing the server protocol is self secure

svaroqui:
Using opensvc and K8S we should use secrets to store ssl certificates and generate service config that expose them via shm to files , this enable to refresh the certificates inside container  without restarting the service but by just replacing the secret  and then use flush ssl or flush privileges

svaroqui:
Add doc Secure cluster in transit explaining all this and may be the janitor proxies

svaroqui:
A new task for DBJobs  Get server certificates 
For replication-manager to work using SSL we have  client certificates config variables .
 ```
--db-servers-tls-client-cert string                    Database TLS client certificate
 --db-servers-tls-client-key string                     Database TLS client key
 
```
Insure that we use those inside configurator instead of nothing  and preserved the one generated, if those client certificates a given to us via config it means it exists some server certificates that we could fetch as well for configurator to integrate and adapt to current database setting  

Document in what context those parameter are used 
And check if they get integrated in configurator seem like a lot redondant and could surelly be simplified  
```
      --db-servers-tls-ca-cert string                        Database TLS authority certificate
      --db-servers-tls-client-cert string                    Database TLS client certificate
      --db-servers-tls-client-key string                     Database TLS client key
      --db-servers-tls-server-cert string                    Database TLS server certificate to push in config
      --db-servers-tls-server-key string                     Database TLS server key to push in config
      --db-servers-tls-use-generated-cert                    Use the auto generated certificates to connect to database backend
      --prov-tls-server-ca string                            server TLS ca
      --prov-tls-server-cert string                          server TLS cert
      --prov-tls-server-key string                           server TLS key
```

svaroqui:
https://dev.mysql.com/doc/refman/8.4/en/using-encrypted-connections.html#using-encrypted-connections-server-side-runtime-configuration

Using MySQL Looks like possible to set the certificates directly in variables 

https://dev.mysql.com/doc/refman/8.4/en/server-system-variables.html#sysvar_auto_generate_certs

PR #894, opened 2024-10-02
auto generate key when init repman

caffeinated92:
auto generate key when init repman

### DIFF
--- a/cluster/srv.go
+++ b/cluster/srv.go
@@ -356,7 +356,7 @@ func (cluster *Cluster) newServerMonitor(url string, user string, pass string, c
 	logInfo, err := os.Stat(errLogFile)
 	if os.IsNotExist(err) || logInfo.Size() > 1024 {
 		// If size is bigger than 1KB when init, rotate it
-		if logInfo.Size() > 1024 {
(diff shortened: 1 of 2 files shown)


svaroqui:
I don't see in this commit where the key is generated  and what happen when it failed for directory grant issue , there is 2 probleme here ,
 1 we could simply consider the repman is run by init scripts or systemd as root and it will success when configpathdir is /etc/replication-manager  but then we need a --user=repman  to switch the pid  user after this step . 
2 we could detect that /etc/replication-manager is not writable and so  use  /home/repman/.replication-manager/ as alternative configpathdir
3 When a root user issue merge-config  or keygen we should check if a key was generated in alternative configpathdir to copy that key in writable  /etc/replication-manager
4 We should state GUI user that the key is not safe in an  alternative configpathdir  and proposed to alert admin for taking action

svaroqui:
PS  /home/repman/ is not hardcoded in the embed binary it could just be the current user when in packages we would have  force repman

svaroqui:
PS 2 This is vital that if 1 is done we tell to guillaume as it change how the rpm and init scripts are deployed

PR #901, opened 2024-10-04
Add validation before using connection for log paths

caffeinated92:
Add validation before using connection for log paths


svaroqui:
You need to backport in 2.3.

PR #906, opened 2024-10-08
Adjust installation script and change it to post-install to minimize chance of failed installation

caffeinated92:
Adjust installation script to chown repman in /var/lib/replication-manager and check for keys instead of authorized_keys


svaroqui:
may be naming it  post-install-script

Issue #915, opened 2024-10-10
Key generation when repman initiated


caffeinated92:
Key generation when repman initiated

caffeinated92:
requested to use /home/repman/.config/replication-manager/etc

PR #916, opened 2024-10-10
Os user scoping and key generation

caffeinated92:
change os user according to --user flag and key generation. Related to #915


caffeinated92:
Currently the result is :
```
INFO[0000] Key not found. Generating : /etc/replication-manager/.replication-manager.key 
INFO[0000] File /etc/replication-manager/.replication-manager.key is not accessible. Try using alternative path: /home/repman/.replication-manager/.replication-manager.key 
INFO[0000] Repman discovered key in alternative path. Using existing key on /home/repman/.replication-manager/.replication-manager.key. 
```

caffeinated92:
requested to use /home/repman/.config/replication-manager/etc

svaroqui:
is this result 
```
INFO[0000] Key not found. Generating : /etc/replication-manager/.replication-manager.key 
INFO[0000] File /etc/replication-manager/.replication-manager.key is not accessible. Try using alternative path: /home/repman/.replication-manager/.replication-manager.key 
INFO[0000] Repman discovered key in alternative path. Using existing key on /home/repman/.replication-manager/.replication-manager.key. 
```
is the output of running repman as root  ?

PR #928, opened 2024-10-18
implement single connection instead of opening new pool

caffeinated92:
implement single connection instead of opening new pool


svaroqui:
We will have to backport  to 2.3

caffeinated92:
Sure. I will backport it to 2.3 branch

PR #930, opened 2024-10-19
Fix reader on master no slave and also fix ONLINE when maintenance 

caffeinated92:
This will fix reader on master no slave and also fix ONLINE when maintenance


caffeinated92:
This should be able to fix condition of #752

PR #1002, opened 2025-01-15
Add API Endpoints for getting attributes using dot notation

caffeinated92:
Add API routes for getting server by index
/api/clusters/{clusterName}/topology/slaves/index/{slaveIndex}

Add API routes for getting attribute by name
/api/clusters/{clusterName}/servers/{serverName}/attr/{attrName}
/api/clusters/{clusterName}/topology/slaves/index/{slaveIndex}/attr/{attrName}


svaroqui:
Can we recheck here the credentials are not exposed

PR #1003, opened 2025-01-15
add BinaryLogName to ServerMonitor and update related methods for binary log handling

caffeinated92:
add BinaryLogName to ServerMonitor and update related methods for binary log handling


svaroqui:
this is no good , what about stopped server they get config on start  this is why i think the variables need to be preserved on disque , and i think we do and reload on repman restart 
Can you check what is done in this regard  and where it is preserved probably srv.Variables

svaroqui:
This is also why i wanted to make it part of a bigger project where we do collect variables based on a local mariadbd pointing to the genreated conf inside the configurator

Issue #981, opened 2025-01-09
Integration of staging clusters  


svaroqui:
A staging cluster is a child cluster of a production cluster containing 3 nodes, one slave is detached from replication and  proxies track the the status of the standalone node to enable dev team to test on production data  

Staging config  important setup :

```
## To be change when proxy tracking state is enable
haproxy = false 

## Specify an other domain for staging 
prov-db-domain = "3" 

##  Specify an other source of replication for the staging cluster
replication-source-name = "staging"

## Point the production cluster to enable mutti-tiers cluster master prod to master staging extra replication source
replication-multisource-head-cluster="prod-cluster"

## Proxy Upgrade script
db-servers-state-change-script = "/data/repman/script/database_state_change.sh"

## To enable ssh to database and proxy hosts, use ssh-keygen ssh-copy-id for first deployment
onpremise-ssh = true
onpremise-ssh-credential = "root:"
onpremise-ssh-private-key = "/root/.ssh/id_rsa_preprod"
scheduler-jobs-ssh = true

##  To automate logical backup  
monitoring-scheduler = true
```

## Lesson learns:  
- you production  and staging monitoring and replication users should have same name same password so that 2 similar users does not get different password in staging after an initial restore from production  
- Mydumper dist version test your restore 

## Todo : 
- New cluster variable --topology-staging bool  
- New cluster variable --topology-staging-refresh-script 
- New cluster variable --topology-staging-post-detach-script ( obfuscation data do whatever have to be done in db)  
- Cluster menu api call to trigger: Staging refresh  

Api call should trigger similar script that will comes embedded in share 
```
#!/bin/bash

# Variables
REPLICATION_MANAGER_USER="admin"
REPLICATION_MANAGER_PASSWORD="xxx"
REPLICATION_MANAGER_URL="https://repman-01:10005"
REPLICATION_MANAGER_CLUSTER_NAME="staging"
REPLICATION_MANAGER_HOST_NAME=$1
REPLICATION_MANAGER_HOST_PORT="3306"
NB_SLAVES=0 

##### This bloc is for getting the replication-manager token 
GET="wget -q --no-check-certificate -O- --header Content-Type:application/json"
AUTH_DATA="{\"username\": \"$REPLICATION_MANAGER_USER\", \"password\": \"$REPLICATION_MANAGER_PASSWORD\"}"
TOKEN=$($GET --post-data "$AUTH_DATA" --header Accept:text/html $REPLICATION_MANAGER_URL/api/login)

[...]

PR #1014, opened 2025-01-20
Implement topology staging

caffeinated92:
Implementing topology staging using script provided in #981


svaroqui (commented):
flags.StringVar(&conf.TopologyStagingRefreshScript, "topology-staging-refresh-script", "", "Topology staging refresh script path. Empty will use copy of embedded template in working directory")

Could this default to the script in share to simplify configuration of the feature to just true false ?

PR #1015, opened 2025-01-21
Feature: drop and rename cluster

### LINKED ISSUE #1008: Remove or rename a cluster when monitoring-save-config is used
It looks that when repman restart it git pull  before push so we can not stop replication rm -rf  datadir/cluster and restart  

- On replication-manager start we should start by a git commit before pull , commit , push , to enable saving  the modifications made on datadir by external user like rm 
- Drop  and Rename  cluster need to to implemented in api and GUI at the cluster list 
- Drop remove all files under the datadir/cluster 
- Rename mv the cluster directory and rename old-cluste-name.toml  to new-cluste-name.toml  replace in all toml [old-cluste-name] and title to new-cluste-name
- Cluster should first be unprovision whatever the orchestrator  so unprovision  ondemand clusters should call and unprovision and mark the cluster with cooky as needProvision
- Following the operation it git commit push the change to gitlab 
- Rename cluster is only possible if the cluster only defined by dynamic config the + button in GUI 
- Rename  cluster have implication on  haproxy  and  config domain_id server_id of nodes

caffeinated92:
Add drop and rename cluster in settings -> general 

- [x] On replication-manager start we should start by a git commit before pull , commit , push , to enable saving the modifications made on datadir by external user like rm
- [x] Drop and Rename cluster need to to implemented in api and GUI at the cluster list
- [x] Drop remove all files under the datadir/cluster
- [x] Rename mv the cluster directory and rename old-cluste-name.toml to new-cluste-name.toml replace in all toml [old-cluste-name] and title to new-cluste-name
- [ ] Cluster should first be unprovision whatever the orchestrator so unprovision ondemand clusters should call and unprovision and mark the cluster with cooky as needProvision
- [ ] Following the operation it git commit push the change to gitlab
- [x] Rename cluster is only possible if the cluster only defined by dynamic config the + button in GUI
- [ ] Rename cluster have implication on haproxy and config domain_id server_id of nodes

### DIFF
--- a/cluster/cluster_acl.go
+++ b/cluster/cluster_acl.go
@@ -22,7 +22,7 @@ type APIUser struct {
 	Password   string          `json:"-"`
 	GitToken   string          `json:"-"`
 	GitUser    string          `json:"-"`
-	IsExternal bool            `json:"-"`
+	IsExternal bool            `json:"isExternal"`
 	Roles      map[string]bool `json:"roles"`
 	Grants     map[string]bool `json:"grants"`
 }
@@ -816,6 +816,9 @@ func (cluster *Cluster) IsURLPassACL(strUser string, URL string, errorPrint bool
 	}
 	if cluster.APIUsers[strUser].Grants[config.GrantClusterDelete] {
 		if strings.Contains(URL, "/api/clusters/actions/delete") {
+			return true
+		}
+		if strings.Contains(URL, "/api/clusters/actions/rename") {
 			return true
 		}
 	}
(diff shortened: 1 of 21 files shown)


caffeinated92:
need explanation for :

- Cluster should first be unprovision whatever the orchestrator so unprovision ondemand clusters should call and unprovision and mark the cluster with cooky as needProvision
- Rename cluster have implication on haproxy and config domain_id server_id of nodes

PR #1071, opened 2025-03-06
Prevent replication-manager to send flapping alert due to transaction inject error

caffeinated92:
Prevent replication manager to send flapping alert due to error while trying transaction inject via proxy


caffeinated92:
Closed due to unclear status

PR #1080, opened 2025-03-31
disable check errant transaction when master state is failed

johnma9380:
## Contributor Agreement

The main fix addresses the issue where, if a failover is triggered while the database GTID is changing rapidly, the GTID of the slave node may become greater than that of the master node. In the current code, only the existence of the check object data in memory is verified, which prevents detecting the fact that the master has crashed. Therefore, an additional state check has been added to the check object data to determine whether the master has already failed.

### DIFF
--- a/cluster/cluster_chk.go
+++ b/cluster/cluster_chk.go
@@ -764,7 +764,7 @@ func (cluster *Cluster) IsSameWsrepUUID() bool {
 }
 
 func (cluster *Cluster) IsNotHavingMySQLErrantTransaction() bool {
-	if cluster.GetMaster() == nil {
+	if cluster.GetMaster() == nil || cluster.GetMaster().State == stateFailed {
 		// disable check if master is crashed as the slave can get more GTID events and so slave GTID is not ubset of masetr GTID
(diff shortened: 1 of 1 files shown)


svaroqui:
Thanks JohnMa sharing same conclusion, but may be this can be improved by removing the master vecteur from the GTID to compare,  select gtid_subset('F36E77A5-0D57-11F0-AD7E-CEACA9FEF976:1-439', 'F36E77A5-0D57-11F0-AD7E-CEACA9FEF976:1-438');
 gtid_subset('F36E77A5-0D57-11F0-AD7E-CEACA9FEF976:1-439', 'F36E77A5-0D57-11F0-AD7E-CEACA9FEF976:1-438') 
0 
would become select gtid_subset('', '');
 1 
for a slave that would be broken by extra write GTID: f36e77a5-0d57-11f0-ad7e-ceaca9fef976:1-439,
f6425674-0d55-11f0-96c2-eee944f76c97:1 
gtid_subset('f6425674-0d55-11f0-96c2-eee944f76c97:1', '') 
1

Issue #1211, opened 2025-09-15
Option for cli passing of username and password from environment variables?


jadestorm:
Hi folk!

Would y'all be willing to include an option for replication-manager-cli to pull it's username and password from an environment variable if it's not specified on the command line?  (maybe something like REPMGR_USERNAME and REPMGR_PASSWORD) I've got a couple of little wrapper scripts I'm putting together for other admins.  (in this particular case, it's just wrapping the docker call so that folk on the host system can just type `replication-manager-cli` and it passes that inside the running docker container.  However, if I user the command line options, that will display the password in plain text in the process list.  It looks like [os.GetEnv](https://pkg.go.dev/os#Getenv) should do the trick.  I could submit a PR but I wasn't sure if you even want this included as a feature.  Thanks!

svaroqui:
Hi Jadestorm, 

We always welcome PR 

Thanks

jadestorm:
No problem!  I'll get y'all a PR then  =)

jadestorm:
Ok I don't know how good or crap that patch is =)  nor am I able to test it yet.  =/  But it's a starting point if nothing else!

svaroqui:
I found your patch  elegant, but it do not solve the global issue with only  2 variables  cover and ENV variables on docker as a strict semantic of being underscore_name  in upper case :   <PRODUCT_NAME>_<VARIABLE_NAME>

jadestorm:
Oh yeah I really like the idea of a more global solution -- just wasn't aware that was an option.  =)  I admit I am not 100% sure where the best place to PUT those calls yet, but let me figure that out please, I would like more golang exposure.  =)

jadestorm:
Hiya @svaroqui   So I finally got to test it and it doesn't seem to be working unfortunately.  =(

I did, however, do some poking around and discovered this snippet/example:

```
    // Define flags
    rootCmd.Flags().String("host", "localhost", "Database host")
    rootCmd.Flags().Int("port", 5432, "Database port")
    rootCmd.Flags().String("user", "admin", "Database user")
    rootCmd.Flags().String("password", "secret", "Database password")

    // Bind flags with Viper
    viper.BindPFlags(rootCmd.Flags())

    // Set environment variable prefix to avoid conflicts with other applications
    viper.SetEnvPrefix("myapp")
    viper.AutomaticEnv()

    // Bind environment variables
    viper.BindEnv("host")
    viper.BindEnv("port")
    viper.BindEnv("user")
[...]

PR #1213, opened 2025-09-16
Add environment variable check for username and password

jadestorm:
This PR is regarding #1211 

While I have a lot of experience with many languages, much to my surprise this is my first time touching golang.  I've tried to make this change as concise as possible, and added a little note to the --help output to reference the env variable option.  I'm not sure I'm happy with the way I worded it.   But if I'm done anything that's considered not a proper way to do things in golang, please be as critical as you wish to be.  =)

I do not currently know how to build a golang app properly.  Are y'all perhaps able to test this or point me at some quick instructions for building it so I can try it out?  If I try a simple go build I get an error.

## Contributor Agreement

### DIFF
--- a/clients/client_cmd.go
+++ b/clients/client_cmd.go
@@ -227,11 +227,10 @@ func cliClusterInServerList() bool {
 }
 
 func getenv(key, fallback string) string {
-    value := os.Getenv(key)
-    if len(value) == 0 {
-        return fallback
+    if value, ok := os.LookupEnv(key); ok {
+        return value
     }
-    return value
+    return fallback
 }
 
 func initServerApiFlags(cmd *cobra.Command) {
--- a/clients/client_cmd.go
+++ b/clients/client_cmd.go
@@ -226,9 +226,17 @@ func cliClusterInServerList() bool {
 	return true
 }
 
+func getenv(key, fallback string) string {
+    value := os.Getenv(key)
+    if len(value) == 0 {
+        return fallback
+    }
+    return value
+}
+
 func initServerApiFlags(cmd *cobra.Command) {
-	cmd.Flags().StringVar(&cliUser, "user", "admin", "User of replication-manager")
-	cmd.Flags().StringVar(&cliPassword, "password", "repman", "Paswword of replication-manager")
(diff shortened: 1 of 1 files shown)


jadestorm:
I did a quick update to use an improved implementation of the env lookup.  I borrowed it verbatum from here:
https://stackoverflow.com/questions/40326540/how-to-assign-default-value-if-env-var-is-empty

svaroqui:
Hello the product use viper 

https://github.com/spf13/viper

So better to follow it  by calling SetEnvPrefix("REPLICATION_MANAGER_CLI")
And later on  AutomaticEnv()

So you get for all cli variables as  ENV variable matches the following format: prefix + "_" + the key name in ALL CAPS.

 in your case REPLICATION_MANAGER_CLI_USER &  REPLICATION_MANAGER_CLI_PASSWORD

svaroqui:
Join us in https://meet.signal18.io via registering a user in https://gitlab.signal18.io , you should receive validation email 
In the meet login page choose the gitlab button and then enter your signal18 gitlab account

svaroqui:
To build we do provide dev docker image with all the necessary tools and script to build  
- only think it is base on the pro release i think so the default should be change in config [default Section} to use prov-orchestrator="onpremise" 
- signal18/replication-manager:dev
- docker run --memory=16g --sysctl net.ipv4.tcp_tw_reuse=1 --sysctl net.core.somaxconn=1024 --sysctl net.ipv4.tcp_fin_timeout=10 --ulimit nofile=262144:262144  -eHOME=/go/src/github.com/signal18/replication-manager --detach --interactive --tty --volume=/etc/localtime:/etc/localtime:ro --volume=/home/ahmad/replication-manager:/go/src/github.com/signal18/replication-manager:rw --volume=/home/ahmad/configs/repman:/etc/replication-manager --volume=/srv/repman-dev2.s18.vol.cloud18/data:/var/lib/replication-manager:rw --volume=/srv/repman-dev2.s18.vol.cloud18/log:/var/log/replication-manager:rw signal18/replication-manager:dev /bin/bash
- adapt your network

jadestorm:
No problem I'll look into making those adjustments and testing it in the dev environment!  Probably will be next week.  Thanks!

jadestorm:
I unfortunately don't have time to test it right now but -- I adjusted things to where I think viper would do the trick.  Does that look sane to you?

PR #1359, opened 2026-02-24
fix: update PrintWithExclude to handle optimizer_switch correctly

caffeinated92:
This pull request introduces a small but important change to the handling of the `optimizer_switch` variable in the `MapValue` printing logic. Specifically, it ensures that whenever `optimizer_switch` is encountered, it is treated as `loose_optimizer_switch` for consistency.

* When printing with exclusion in the `MapValue.PrintWithExclude` method, the code now checks if the `varname` is `optimizer_switch` and, if so, changes it to `loose_optimizer_switch` before proceeding.


caffeinated92:

PR #1363, opened 2026-02-27
fix: enhance splitdump restore logic to handle GTID and missing table scenarios

caffeinated92:
- Summary:
  - Skip mysql.gtid_slave_pos data files during splitdump restores and add warning logs for skipped files.
  - Add restore preamble to set FOREIGN_KEY_CHECKS and schema context before loading data.
  - Gracefully ignore missing mysql system tables during restore to avoid hard failures.
  - Add filename parsing helpers and tests for schema/table detection and error classification.


caffeinated92:

PR #1364, opened 2026-02-27
fix: prevent splitdump restore from binlogging

caffeinated92:
Apply per-file sql_log_bin preambles on server restores and add a CLI flag to disable binlogging, with tests for preamble generation.


caffeinated92:

PR #1367, opened 2026-02-27
fix: harden splitdump sharding and restore file coverage

caffeinated92:
- Summary:
  - Defer shard creation until the end of an SQL statement to avoid tuple‑only shards and keep each shard self‑contained.
  - Reset shard numbering per table and require active table context before splitting to prevent incorrect filenames.
  - Expand splitdump tests for boundary conditions, missing USE, alternating shard/no‑shard tables, and CLI stream‑size validation.
- Proof:
  - Verified with mysqldump.sql.gz and 1 MiB shards; shard headers show USE <schema>; followed by complete statements.
- Tests:
  - go test ./utils/splitdump/...
  - go test -tags clients ./clients/...

### DIFF
--- a/clients/client_splitdump_test.go
+++ b/clients/client_splitdump_test.go
@@ -6,6 +6,7 @@ package clients
 import (
 	"os"
 	"path/filepath"
+	"strings"
 	"testing"
 )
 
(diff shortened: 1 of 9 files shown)


caffeinated92:
Addressed review items (commit 5886b2a03):
- Data race in TestSplitDumpLineParserNoShardWhenLimitEqualsSize  
  streamSizeMax is computed before go SplitDumpLineParser(...) starts, so there’s no shared mutation after the goroutine begins.
- Missing newline in shard log  
  fmt.Printf("Processing table data %s\n", tableName) now uses a newline for consistent logging.
- splitReady defensive reset  
  Added an else { splitReady = false } to avoid stale state if input deviates.
- isStatementEnd comment  
  Added a short note explaining the ; suffix check is sufficient for mysqldump’s default output.

PR #1373, opened 2026-03-03
Enable safe deletion of local backups while guarding active jobs

caffeinated92:
Summary
This PR adds a first-class delete path for non‑restic backups, wires it into the Maintenance UI, and hardens the backend to prevent deleting a backup while its job is still running. It keeps restic snapshots untouched and focuses on reclaiming local disk space safely.
What changed
- Backend delete API: Adds /api/clusters/{clusterName}/backups/{backupID}/delete with ACL enforcement and JSON response.  
- Cluster delete logic: Deletes the backup payload and metadata file, with a strict path‑within‑backup‑directory guard.  
- Active job guard: Prevents deleting a backup that matches the currently running backup metadata.  
  - Uses BackupSessionID/ID/dest matching.
  - Treats unknown backup tools as “check both logical/physical” to avoid false negatives.
  - Rechecks before deletion to reduce TOCTOU windows.
- Client + UI: Adds delete action in “Current Backups” with confirm prompt and auto‑refresh of list/stats after success.  
- Restic behavior: Deleting local backups does not purge restic snapshots.
Why
- Operators need to free local disk space even after a restic snapshot exists.
- Deleting the wrong file during an active backup can corrupt the running job; the new guard blocks that safely.
Behavior notes
- If a matching backup is still running, the API returns 409 Conflict and deletion is blocked.
- Restic snapshots are managed by their own purge flow and are not affected by delete‑backup.

### DIFF
--- a/cluster/backup_helpers.go
+++ b/cluster/backup_helpers.go
@@ -8,6 +8,7 @@ package cluster
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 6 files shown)


caffeinated92:
@claude review 9f72fb24475d62125446b7445fda9f9cc4b4b705

PR #1375, opened 2026-03-04
fix: replace restic busy-wait recursion with cond-based lock

caffeinated92:
**Summary**
- Replace the recursive sleep/poll gate around restic operations with a condition‑based lock to avoid unbounded recursion and reduce contention.
- Make FetchRepo use the same lock+timeout path for consistent behavior across restic operations.

**Notes / Follow-ups**
- GetCanFetch/SetCanFetch are not mutex‑protected and now become more likely to race under high concurrency; consider switching to sync/atomic or guarding under the existing mutex.
- PurgeRepo previously busy‑waited with sleep+recurse when CanFetch == false; this was a pre‑existing design concern addressed here by the lock.
- The immediate purge path (NeedPurgeNow) still skips FetchRepo while queued PurgeTask triggers it; this is unchanged but worth documenting as intentional behavior.

### DIFF
--- a/utils/backupmgr/restic.go
+++ b/utils/backupmgr/restic.go
@@ -809,7 +809,7 @@ func (repo *ResticManager) worker() {
 		}
 
 		if repo.NeedPurgeNow {
-			// Process purge now without holding the mutex.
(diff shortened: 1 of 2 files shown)


caffeinated92:

PR #1385, opened 2026-03-05
fix: use correct TLS config name in mysql.RegisterTLSConfig

stefnats:
## Summary
- Fix TLS config name mismatch in `mysql.RegisterTLSConfig()` that silently disables TLS
- `RegisterTLSConfig("&tls=tlsconfig", ...)` registers under wrong name — driver looks up `"tlsconfig"` from DSN but can't find it
- Splits `ConstTLSCurrentConfig`/`ConstTLSOldConfig` into separate DSN fragment and registration name constants

## Impact
- **All MySQL 8.4+ users affected** — `caching_sha2_password` (default auth) requires TLS for initial handshake
- Without this fix, repman silently connects without TLS, causing `Access denied` or `Authentication requires secure connection` errors
- Workaround: switch users to `mysql_native_password` + `REQUIRE NONE`

## Test plan
- [ ] Connect repman to MySQL 8.4 with `caching_sha2_password` + `REQUIRE SSL` users
- [ ] Verify `SHOW STATUS LIKE 'Ssl_cipher'` shows active TLS on repman's connection
- [ ] Verify failover/switchover works with TLS-required users

### DIFF
--- a/cluster/srv.go
+++ b/cluster/srv.go
@@ -287,6 +287,8 @@ const (
 	ConstTLSNoConfig      string = ""
 	ConstTLSOldConfig     string = "&tls=tlsconfigold"
 	ConstTLSCurrentConfig string = "&tls=tlsconfig"
+    ConstTLSCurrentConfigName string = "tlsconfig"
+    ConstTLSOldConfigName     string = "tlsconfigold"
 )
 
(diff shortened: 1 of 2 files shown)


tanji:
Thanks for the fix!

PR #1387, opened 2026-03-05
fix(cluster): reduce schema checksum work without shardproxy

caffeinated92:
**Summary**
- gate schema checksum/replica scans behind a `checksum` flag so non-shardproxy runs only refresh table metadata
- keep shardproxy duplicate detection and vtable creation tied to checksum-enabled runs
- align on-demand dict table refresh and scheduler trigger with the new checksum-aware schema monitor


caffeinated92:

PR #1399, opened 2026-03-11
fix(restic): allow admin-selected mount dirs under strict checks

caffeinated92:
Accept mount paths that resolve to the base directory while still rejecting path escapes, so admins can use explicit absolute mount directories. Add tests to lock the relaxed behavior and keep traversal protections in place.


tanji:
@claude

caffeinated92:
@claude

PR #1400, opened 2026-03-11
 feat: backup encryption

caffeinated92:
## Summary
- add AES-256-CBC backup encryption with CLI decrypt command and secret-aware passphrase resolution
- support directory logical backups via archive encryption (tar/tar.gz) with streaming OpenSSL encrypt/decrypt and restic compatibility
- introduce per-file directory encryption mode (opt-in) with `encryptionMode`/`encryptionTool` metadata for forward-compatible restores
- add OpenSSL startup observability and hard-fail when encryption is enabled but OpenSSL is missing
- update implementation docs covering phase 1 and phase 2 encryption behavior
## Testing
- go test ./cluster/... ./server/... ./config/...


caffeinated92:
@claude

caffeinated92:
I found 3 important issues to address before merge:
- High Encrypted single-file logical backups are not decrypted on restore
  - encryptBackupMetadataFile marks encrypted files with EncryptionMode = "archive" (cluster/srv_job_backup.go:4600).
  - But restore mode detection only treats "per-file" as encrypted metadata; otherwise it only infers archive mode from .tar.enc/.tar.gz.enc suffix (cluster/srv_job_backup.go:5294).
  - For common logical file backups like *.sql.gz.enc, prepareEncryptedDirectoryLogicalRestorePath returns the original encrypted path unchanged (cluster/srv_job_backup.go:5312), so mysqldump reseed paths get ciphertext.
  - This likely breaks restore for encrypted non-directory logical backups.
- High Tar extraction still allows symlink write-chain traversal for regular files
  - In extractArchiveToDir, regular entries are written directly to destAbs after lexical path checks (cluster/srv_job_backup.go:5139), but there’s no symlink-resolution check on parent or final path.
  - If an earlier archive entry creates a symlink inside tree, a later regular file entry targeting that path can escape the extraction root.
  - This is exactly the threat modeled by TestExtractArchiveToDirBlocksSymlinkTraversalWriteChain (cluster/srv_job_backup_extract_test.go:118), but current extraction logic doesn’t appear to enforce it for tar.TypeReg.
- Medium CLI decrypt still exposes passphrase via process arguments
  - decryptWithPassphraseOpenSSL builds OpenSSL args with -pass pass:<password> (clients/client_decrypt_backup.go:110).
  - That leaks secret material to process list/audit surfaces, unlike the server path that uses fd:3.
  - Suggest aligning CLI decryption to fd-based passphrase handling for consistency and safer local ops.

PR #1419, opened 2026-03-31
fix: reduce stale Database Jobs reads in dashboard and cluster API

caffeinated92:
## Summary
- refresh Maintenance job data immediately when entering the tab instead of waiting for the next poll
- prevent cross-cluster job response bleed in the React store
- refresh cluster job entries on API reads when scheduler-driven invokers lag, while only merging newer DB task state into cache
## Why
Database Jobs could appear stale even when the underlying job state had already advanced. The issue was a mix of delayed frontend fetch timing and server-side cache refresh behavior that depended too heavily on background invokers.
## Changes
- Dashboard:
  - trigger immediate Maintenance data fetch on tab entry
  - scope `getJobs` request de-duplication per cluster
  - guard Redux job updates so only the active cluster can overwrite job state
- Cluster:
  - refresh job entries when cache is empty, marked dirty, or scheduler monitoring is disabled
  - update cached job records from DB only when the DB row is newer, with same-timestamp field-change fallback
## Impact
- reduces stale Job views in Maintenance
- improves correctness when switching clusters
- makes cluster job entry reads more resilient when scheduler update timing lags

### DIFF
--- a/cluster/cluster_job.go
+++ b/cluster/cluster_job.go
@@ -24,6 +24,8 @@ import (
 	"github.com/signal18/replication-manager/utils/state"
 )
 
+const jobsAPIDBRefreshMinInterval = 5 * time.Second
+
 func (cluster *Cluster) JobAnalyzeSQL(persistent bool) error {
 	var err error
 	var logs string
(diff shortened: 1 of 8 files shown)


caffeinated92:
@claude review

PR #1423, opened 2026-04-01
fix(terminal): Support OpenSVC container#jobs terminal

caffeinated92:
Summary
- add OpenSVC web terminal support for selecting container#jobs in addition to the default container#db
- validate rid server-side so it is only accepted for OpenSVC server bash terminals, with tests for valid and invalid paths
- fix terminal UI websocket state handling so stale socket events no longer show disconnected while the active session is connected


caffeinated92:
@claude review

caffeinated92:
@claude review

PR #1432, opened 2026-04-06
fix(config/manager): prevent shutdown deadlocks and harden go-git metadata recovery

caffeinated92:
## Summary
This PR hardens `config/manager` around failure seen in production:
- go-git push failures such as `object not found` could leave the local repo metadata in a bad state, previously requiring manual `.git` removal to recover
The changes make shutdown behavior deterministic, prevent new work from being queued once shutdown begins, and replace destructive `.git` deletion with a safer metadata refresh flow that preserves dirty working-tree files in place.

### Shutdown / queue handling
- drain queued save tasks and queued git tasks when stop is observed
- release waiting callers before worker exit instead of spinning on closed stop channels
- prevent new save/push/pull tasks from being enqueued once shutdown has started
- make shutdown state checks concurrency-safe with atomic stop flags
### Commit manager hardening
- synchronize `CommitManager.Stop()` with the commit queue
- clear pending commit tasks under lock before releasing waiters
- guard stop-channel close and make stop-state reads/writes safe across goroutines
### go-git recovery
- detect recoverable push failures including missing object/reference style errors
- replace `os.RemoveAll(<workdir>/.git)` recovery with in-place metadata refresh
- clone fresh repo metadata into a temp directory, then swap only `.git`
- preserve dirty working-tree files in place for running processes
- keep rollback behavior if metadata swap fails
- remove the hardcoded `master` branch assumption during clone setup

## Why
We observed intermittent go-git errors like `cannot push: object not found`, where the practical recovery had been deleting `.git` and recloning. That approach is unsafe when the working tree contains dirty files or is actively used by other processes.
We also identified shutdown paths where workers could stop with queued tasks still present, leaving waiters blocked indefinitely. This PR makes shutdown best-effort and non-blocking by draining queued work and exiting cleanly.

## Behavior impact
- active work still follows the existing save/push sequencing
- git metadata refresh now preserves local files instead of deleting `.git` outright

## Validation
- `go test ./config/manager`
- `go test ./config/...`

### DIFF
--- a/cluster/cluster.go
+++ b/cluster/cluster.go
@@ -32,6 +32,7 @@ import (
 
 	"github.com/pelletier/go-toml"
 	"github.com/signal18/replication-manager/cluster/configurator"
+	"github.com/signal18/replication-manager/cluster/logplugin"
 	"github.com/signal18/replication-manager/cluster/nbc"
 	"github.com/signal18/replication-manager/config"
 	"github.com/signal18/replication-manager/config/manager"
@@ -52,7 +53,6 @@ import (
 	log "github.com/sirupsen/logrus"
 	logsql "github.com/sirupsen/logrus"
 	"golang.org/x/crypto/ssh"
-	"github.com/signal18/replication-manager/cluster/logplugin"
 )
 
 var clusterError = config.ClusterError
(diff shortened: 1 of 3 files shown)


caffeinated92:
@claude review

caffeinated92:
Yes — these two follow-up commits materially improve the PR, and the second one is the one that answers the likely reviewer concern.

What each commit fixed:

2851179

hardens shutdown/queue behavior with a sync.Cond instead of polling
protects clusterData with clusterMu
makes SaveConfig(..., true) return promptly during stop/race cases
improves panic recovery/logging and adds regression tests for those paths

So this commit is mostly about the concurrency / deadlock side of the PR, not the branch issue.

a1fa946

adds resolveCurrentLocalBranch() and uses the current local branch when refreshing metadata
if clone-by-reference fails, it retries with the remote default branch
restores ConfRestoreOnStart checkout behavior
makes .git swap robust for cross-filesystem rename failures (EXDEV) via copy fallback
adds a test proving RefreshGitMetadata preserves the active branch reference (feature in the test)

That means the original reviewer concern about “hardcoded master” is substantially addressed by a1fa946, because refresh now prefers the active branch and has a fallback retry path instead of only assuming one branch.

The one caveat: there is still a master fallback inside resolveCurrentLocalBranch() / RefreshGitMetadata when local branch detection fails. It is no longer the primary behavior, but it still exists as a fallback path.

So my read is:

if the reviewer complained that refresh was still tied to master, your second follow-up commit mostly fixes it
if the reviewer is very strict, they may still point out that master is not fully eliminated, only downgraded to a fallback path

PR #1434, opened 2026-04-07
fix(server): expose git push/pull failures through global state-machine alerts

caffeinated92:
## Summary
- Add git supervision state production to the server global state-machine so git sync problems are visible via global alerts.
- Track push and pull outcomes in `ConfigManager` with a thread-safe health snapshot, then map warning/error classes to `GWARN/GERR` states.
- Add global error templates and tests covering warning/error emission, push vs pull separation, and alert resolution after successful sync.

## Why
Git sync failures were only logged, which made them easy to miss operationally.  
This change surfaces those failures as first-class global alerts so operators can detect and react faster.


caffeinated92:
@claude review

PR #1439, opened 2026-04-10
Harden app monitor creation validation and improve New Server modal defaults

caffeinated92:
### **Summary**
- improve the New Server modal UX for app monitors with clearer validation, name-to-host autofill, and config-based default ports for non-app types
- make config-driven DB/proxy ports read-only in the modal to better reflect backend behavior
- harden backend app creation validation so invalid host, port, and registry inputs are rejected before initiation

### **Why**
App initiation was working, but the frontend and backend allowed a few ambiguous or unsafe cases:
- the modal could submit empty or misleading port values
- app setup lacked a dedicated helper flow for name/host entry
- DB/proxy/app behavior was mixed together in one form, which made it easier to misconfigure fields
- backend app creation relied too heavily on frontend correctness and did not explicitly reject some invalid inputs
These changes reduce operator error and make the form behavior match how the backend actually treats each monitor type, while preserving the current initiation flow for valid requests

### **Details**
**Frontend**
- add app-specific helper input for name and auto-fill host from it
- keep host editable for app monitors
- require valid ports in the modal
- auto-fill DB/proxy ports from cluster config/backend defaults
- make non-app ports read-only because they are config-driven
- improve field validation and error display
- improve modal styling and clarity around template/port behavior

**Backend**
- trim and validate app add inputs before initiation
- reject empty app host
- reject invalid app port values
- require docker image or template
- validate private registry input more explicitly
- keep template-based fallback behavior intact so valid app initiation still works as before

### **Notes**
- no backend contract change was introduced for app creation
- valid existing app initiation paths are preserved
- stricter validation only blocks clearly invalid requests

### DIFF
--- a/cluster/cluster_app.go
+++ b/cluster/cluster_app.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"path/filepath"
 	"runtime"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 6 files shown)


caffeinated92:
@claude review

PR #1445, opened 2026-04-14
fix splitdump restore ordering with manifest and FK-aware fallback

caffeinated92:
Splitdump restore could fail when child table schemas were restored before parent tables (alphabetical ordering reintroduced an old FK ordering bug).

This change:
- Adds a manifest file to splitdump backups that records emission order
- Uses manifest order directly during restore when present and valid
- Falls back to FK dependency analysis for old backups without manifest (parent tables are restored before child tables)
- Adds transparent retry on bufio.ErrTooLong via semicolon-delimited statement parsing for oversized DDL files
- Keeps existing mysql.proc overlap pruning intact

Includes regression tests for manifest emission, manifest restore ordering, FK-based ordering, and oversized plain/gzip DDL retry.


caffeinated92:
@claude review

PR #1446, opened 2026-04-14
fix: change RelayLogSpace from sql.NullInt64 to sql.Null[uint64]

### LINKED ISSUE #1442: [Bug]: SHOW REPLICA STATUS parsing fails with uint64 overflow on Relay_Log_Space (Percona 8.4)
### Component

Server (monitoring/orchestration)

### Bug Description

Replication Manager fails to parse SHOW REPLICA STATUS when the Relay_Log_Space column returns a large uint64 value that exceeds int64 range. This causes repman to lose visibility of the slave in the topology, leading to cascading failures: the slave is seen as StandAlone, no failover candidates are available, and if the master subsequently fails, the resulting automatic failover creates a split-brain situation.

The specific error:

sql: Scan error on column index 22, name "Relay_Log_Space": converting driver.Value type uint64 ("18446744073709442261") to a int64: value out of range
The value 18446744073709442261 is near uint64 max (18446744073709551615), suggesting Percona 8.4 returns a uint64 for this column while repman's Go code scans it into an int64.

### Steps to Reproduce

1. Configure replication-manager to monitor a Percona Server 8.4 async GTID cluster (master + slave)
2. Run the cluster under write load until relay logs grow large or Relay_Log_Space reports a large value
3. Observe monitoring logs

### Expected Behavior

Replication Manager should correctly parse SHOW REPLICA STATUS regardless of the Relay_Log_Space value, and continue to recognize the slave in the topology.

### Actual Behavior

Repman logs the scan error and then reports:

- Could not find a slave in topology (ERR00010)
- No replica found for routing reads (ERR00085)
- No candidates found in slaves list (ERR00032)
The slave node is effectively invisible to repman. If the master subsequently fails, repman attempts failover but the slave was already operating independently (writes were going to it via ProxySQL reader pool without replication). This creates a split-brain with millions of errant GTIDs on both nodes.

The error occurs intermittently (not every monitoring tick), which makes it harder to notice.

### Relevant Logs

```shell
Apr 09 20:23:52 replication-manager: level=error msg="Could not get slaves status 10.0.0.1:3306 sql: Scan error on column index 22, name \"Relay_Log_Space\": converting driver.Value type uint64 (\"18446744073709442261\") to a int64: value out of range" cluster=my-cluster module=Monitor server="10.0.0.1:3306" sql="SHOW REPLICA STATUS"

Apr 09 20:23:54 replication-manager: level=warning msg="Could not find a slave in topology" cluster=my-cluster code=ERR00010
Apr 09 20:23:54 replication-manager: level=warning msg="No replica found for routing reads" cluster=my-cluster code=ERR00085
Apr 09 20:23:54 replication-manager: level=warning msg="No candidates found in slaves list" cluster=my-cluster code=ERR00032

# Same error repeated at ~1-3 hour intervals:
Apr 09 21:00:02 ... uint64 ("18446744073708974793") ...
Apr 09 22:14:09 ... uint64 ("18446744073708722727") ...
Apr 09 23:16:36 ... uint64 ("18446744073708578215") ...
Apr 10 00:47:22 ... uint64 ("18446744073707568971") ...

# Eventually master (10.0.0.2) went down, failover triggered, split-brain:
Apr 10 02:25:10 ... level=error msg="Server 10.0.0.2:3306 state changed from Suspect to Failed"
Apr 10 06:00:03 ... level=warning msg="Extra master in master-slave topology, rejoining after split-brain" code=ERR00063
```

### Replication Manager Version

v3.1.22 (Build Time: 2026-03-12)

### Database System

Percona Server

### Database Version

8.4.0-1

### Replication Mode

GTID-based (standard)

### Proxy Type

ProxySQL

### Operating System

Ubuntu 24.04.3 LTS

### Configuration

```toml
[my-cluster]
db-servers-hosts               = "10.0.0.1:3306,10.0.0.2:3306"
db-servers-prefered-master     = "10.0.0.1:3306"
failover-mode                  = "automatic"
failover-limit                 = 5
failover-time-limit            = 3600
failover-falsepositive-ping-counter = 5
failover-falsepositive-heartbeat    = true
autorejoin                     = true
autorejoin-script              = "/usr/local/bin/repman-autorejoin.sh"
proxysql                       = true
replication-use-ssl            = true
```

### Additional Context

The root cause appears to be in the Go database driver scanning Relay_Log_Space into int64 when Percona 8.4 returns uint64. The fix would be to use uint64 or sql.NullInt64 with overflow handling for this column.

This is particularly dangerous because the monitoring failure is silent (no alert triggered) and intermittent, so the topology degrades gradually. When a real failover is needed, the cluster is already in a broken state.

A secondary issue: the autorejoin-script receives the rejoining node after a split-brain with potentially millions of errant GTIDs. The built-in empty transaction injection approach (one transaction per GTID) is too slow for large errant sets — in our case 2.1M errant GTIDs would have taken ~5 days at 1 GTID/mysql-invocation.

caffeinated92:
Percona/MySQL 8.4 can return Relay_Log_Space as a large uint64 that exceeds int64 range, causing SHOW REPLICA STATUS parsing to fail with:
  sql: Scan error on column index 22, name \"Relay_Log_Space\":
  converting driver.Value type uint64 to a int64: value out of range

This scan failure causes the replica to disappear from topology detection, leading to cascading failures:
- slave seen as StandAlone
- no failover candidates available
- automatic failover can create split-brain

The fix changes the field type to sql.Null[uint64] which safely handles large unsigned values and preserves NULL semantics.

Also adds comprehensive tests covering:
- NULL, uint64, int64, MaxUint64, negative rejection
- []byte and string numeric inputs
- overflow and non-numeric rejection

### DIFF
--- a/utils/dbhelper/replication_test.go
+++ b/utils/dbhelper/replication_test.go
@@ -8,6 +8,7 @@
 package dbhelper
 
 import (
+	"math"
 	"strings"
 	"testing"
 )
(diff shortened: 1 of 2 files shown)


stefnats:
Thanks for the quick fix! The change looks clean and correct.

We can confirm this fixes the exact issue we hit in production. Percona Server 8.4.0-1 was returning `Relay_Log_Space` values like `18446744073709442261` (uint64 underflow from [MySQL Bug #87832](https://bugs.mysql.com/bug.php?id=87832) regression), which caused the `sql.NullInt64` scan to fail and made our replica invisible to repman, ultimately leading to a split-brain after automatic failover.

Happy to test a build with this patch on our cluster if that helps.

caffeinated92:
Actually we can just remove it from list, since it's not consumed. But was unsure about it. At least this will catch it correctly with uint64

svaroqui:
> Thanks for the quick fix! The change looks clean and correct.
> 
> We can confirm this fixes the exact issue we hit in production. Percona Server 8.4.0-1 was returning `Relay_Log_Space` values like `18446744073709442261` (uint64 underflow from [MySQL Bug #87832](https://bugs.mysql.com/bug.php?id=87832) regression), which caused the `sql.NullInt64` scan to fail and made our replica invisible to repman, ultimately leading to a split-brain after automatic failover.
> 
> Happy to test a build with this patch on our cluster if that helps.

You can use  our nightly docker images if needed  they are build for every commit in develop  also i have new PR on security and workload  plugins , are you intrested in testing those plugins  ? I have only tested them over MariaDB so far  but they should not impact anything if not working only  making false positive logs

PR #1447, opened 2026-04-14
fix: update React dashboard to resolve Vite 8 peer dependency conflict

caffeinated92:
The dashboard build was failing with:
- npm ERR! ERESOLVE override for @vitejs/plugin-basic-ssl
- sh: vite: not found (due to failed install)

Root cause: vite@8.0.7 is incompatible with @vitejs/plugin-basic-ssl@1.2.0 and @vitejs/plugin-react@4.3.4 which only support up to Vite 6.

Changes:
- Downgrade vite from ^8.0.7 to ^6.0.0 (compatible with existing plugins)
- Fix SCSS import resolution in vite.config.js (path from ./src to /src)

Fixes CI/Docker builds that run "make react" or "make pro cli"

### DIFF
--- a/share/dashboard_react/package-lock.json
+++ b/share/dashboard_react/package-lock.json
@@ -55,7 +55,7 @@
         "eslint-plugin-react-hooks": "^4.6.2",
         "eslint-plugin-react-refresh": "^0.4.19",
         "prettier": "^3.5.1",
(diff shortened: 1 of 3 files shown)


caffeinated92:
We will bump vite later

PR #1451, opened 2026-04-14
fix: unlock restic repo at startup to clear stale locks

caffeinated92:
The change proactively clears stale Restic locks on startup (e.g., after crashes), which is a reliability bugfix.


caffeinated92:
@claude review

PR #1455, opened 2026-04-20
fix: improve graceful shutdown by refactoring git mutex handling

caffeinated92:
Release gitMutex before git push/pull network operations to prevent blocking ConfigManager.Stop() during process termination. Add stop-channel check to abort in-flight saves on shutdown request.


caffeinated92:
@claude review

PR #1457, opened 2026-04-21
OpenSVC v3 provisioning and fix optimizer_switch boot compatibility

caffeinated92:
## Summary
This PR improves OpenSVC v3 compatibility and provisioning reliability across database, app, and proxy services.
### What changed
- Added/expanded OpenSVC v3 handling for app/proxy lifecycle operations (provision, unprovision, start/stop/restart).
- Split template generation into reusable section-map builders and introduced INI template rendering for v3 flows.
- Updated v3 provisioning flow to:
  - create template,
  - handle existing-object conflicts (HTTP 409) safely,
  - then explicitly trigger provision.
- Introduced structured OpenSVC `StatusError` handling and improved create/update behavior for config/secret key-values.
- Wipe `02_delta.cnf` before provisioning/init to avoid stale runtime delta carrying into new provision cycles.
- Fixed optimizer boot issues by outputting `loose_optimizer_switch` where `optimizer_switch` options are emitted.
- Updated OpenSVC bootstrap script to safely handle missing `/bootstrap/version` and ensure `/bootstrap` exists early.
## Why
OpenSVC v3 has different API behavior and stricter object lifecycle semantics than earlier flows.  
These changes make provisioning more idempotent and resilient (especially on retries/re-provision), while also preventing startup failures caused by strict optimizer switch parsing.
## Impact
- Safer reprovision behavior when objects/templates already exist.
- Better alignment between generated templates and OpenSVC v3 expectations.
- Reduced risk of failed DB startup due to optimizer switch option handling.
- Cleaner provisioning state by removing stale delta config before re-init/re-provision.

### DIFF
--- a/cluster/cluster_acl.go
+++ b/cluster/cluster_acl.go
@@ -411,6 +411,24 @@ func (cluster *Cluster) IsURLPassACL(strUser string, URL string, errorPrint bool
 }
 
 func (cluster *Cluster) IsURLPassAppsACL(strUser string, URL string) bool {
+	// App settings actions must only be evaluated against app settings rules.
+	// This prevents fallback from /settings/actions/* to generic /actions/* rules.
+	if strings.Contains(URL, "/settings/actions/") {
+		settingsRules := make([]ACLRule, 0)
+		for _, rule := range appACLRules {
+			if strings.Contains(rule.URLPattern, "/settings/actions/") {
+				settingsRules = append(settingsRules, rule)
+			}
+		}
+
+		if cluster.matchACLRules(strUser, URL, settingsRules) {
+			return true
+		}
+
+		cluster.LogModulePrintf(cluster.Conf.Verbose, config.ConstLogModGeneral, config.LvlInfo, "ACL app settings check failed for user %s : %s", strUser, URL)
+		return false
+	}
+
 	// Check against rule-based ACL system with improved logging
 	if cluster.matchACLRules(strUser, URL, appACLRules) {
 		return true
--- a/cluster/cluster_acl_rules.go
+++ b/cluster/cluster_acl_rules.go
@@ -29,6 +29,8 @@ var databaseACLRules = []ACLRule{
 	{"/actions/unprovision", nil, []string{config.GrantProvDBUnprovision}},
 	{"/actions/start", nil, []string{config.GrantDBStart}},
 	{"/actions/stop", nil, []string{config.GrantDBStop}},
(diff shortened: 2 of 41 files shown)


caffeinated92:
@claude review de44820

caffeinated92:
@claude review

caffeinated92:
I reviewed PR #1457 (read-only) via gh and the current diff looks much better than earlier iterations.
Verdict
- No critical/high blockers found
- 1 medium issue still worth fixing before merge
- A couple of low/nit items for consistency
---
Medium
1) Auto-reload pause can be released too early when Menu + ConfirmModal overlap
- share/dashboard_react/src/components/MenuOptions/index.jsx:27-40, 51-56
- share/dashboard_react/src/components/Modals/ConfirmModal/index.jsx:42-55, 66-71
Both components maintain separate counters but share one pause_auto_reload key with token ownership checks.  
If one component opened first and the other opens while pause is already active, the second won’t set the key. When the first closes, it may clear pause_auto_reload even though the second is still open.
Suggested fix: centralize pause handling with a single shared refcount utility:
- acquirePause() increments global count; sets key on transition 0 -> 1
- releasePause() decrements global count; removes key on transition 1 -> 0
---
Low / consistency
2) App handlers still return “Server Not Found”
- server/api_app.go:709, 758
- In app-specific abort/clear handlers, response text should ideally be "App Not Found" for API consistency (similar cleanup done for proxy).
3) Behavioral change in WriteDeltaVariables is intentional but should be called out in PR description
- Runtime fallback path was removed; now deployed-only delta lines are written.
- Not a blocker, but worth explicitly documenting as monitoring behavior change.
---
What looks good
- OpenSVC v3 abort/clear routing + ACL mapping is coherent.
- Unprovision channel handling is significantly safer.
- Proxy SetTemplateMD5 interface fix is clean.
- ProvObjectAllowOverwrite handling restored in v3 map creation.
- Template payload decode compatibility (string/object) is robust.
- Tests added in the right areas (prov_opensvc_v3_test, ACL tests, config tests).

caffeinated92:
Review result
- ✅ No critical/high issues found
- ✅ App error message fix (App Not Found) is correct
- ✅ Shared pause lock utility fixes the menu/modal overlap race condition
- ✅ Pause key restore behavior is consistent for pre-existing pause state
Validation status
- ✅ go test ./cluster ./server ./config ./opensvc -count=1 passed
- ✅ npm run build passed

PR #1458, opened 2026-04-22
implement missing run task v3

caffeinated92:
Implement Run Task for opensvc V3 and set orphan message as debug to reduce flooding


caffeinated92:
@claude review

PR #1474, opened 2026-05-07
Add OpenSVC runtime pools endpoint and fail-fast app storage

caffeinated92:
## Summary
- add OpenSVC runtime pool discovery for v2/v3 with shared-capability parsing and comprehensive tests
- add protected `GET /api/clusters/{clusterName}/opensvc-pools` and wire React storage UI to consume dynamic pool options
- remove static app pool config dependency, enforce explicit/fail-fast app volume selection, and set shared volume flag from runtime pool metadata
- add pool-info caching at cluster level (5-minute cached reads) with provisioning path using forced fresh fetch

## Validation
- `go test ./opensvc ./cluster ./server`
- React targeted storage/Redux flow validated in prior runs

## Notes
- branch includes ACL updates for `/opensvc-pools` and UI filtering for shared pools in failover topology

### DIFF
--- a/cluster/cluster.go
+++ b/cluster/cluster.go
@@ -308,6 +308,9 @@ type Cluster struct {
 	OrchestratorVersion   string       `json:"-"`
 	orchestratorVersionMu sync.RWMutex `json:"-"`
 	lastOrchestratorProbe time.Time    `json:"-"`
+	opensvcPoolInfoCache   []opensvc.PoolInfo `json:"-"`
+	opensvcPoolInfoCacheAt time.Time          `json:"-"`
+	opensvcPoolInfoCacheMu sync.RWMutex       `json:"-"`
 	// Per-cluster preserved variables (replaces ProvDBConfigPreserveVars mechanism)
 	preservedVars               map[string]string          `json:"-"`
 	preservedVarsExcludeServers map[string]map[string]bool `json:"-"` // varName -> {serverID -> true}
--- a/cluster/cluster_acl_rules.go
+++ b/cluster/cluster_acl_rules.go
@@ -239,6 +239,7 @@ var clusterACLRules = []ACLRule{
 	{"/services/actions/unprovision", nil, []string{config.GrantProvClusterUnprovision}},
 	{"/api/clusters/actions/add", nil, []string{config.GrantProvCluster, config.GrantClusterCreate}},
 	{"/opensvc-gateway", nil, []string{config.GrantProvCluster}},
+	{"/opensvc-pools", nil, []string{config.GrantAppDeployment, config.GrantProvCluster}},
 
 	// Staging
 	{"/actions/staging-refresh", nil, []string{config.GrantClusterStaging}},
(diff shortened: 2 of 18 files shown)


caffeinated92:
Looks good to me overall. I think this PR is merge-worthy.

The main direction makes sense: app volume pool selection should come from the OpenSVC runtime state instead of the static `prov-app-volume-pools` config. Removing the old config-based pool parsing and making app storage fail fast when no explicit volume/pool exists is a good improvement. It avoids silently creating/selecting a local volume with assumptions that may not match the actual OpenSVC environment.

Backend changes look coherent:
- The new `opensvc-pools` endpoint is placed in the protected cluster API path.
- ACL wiring looks appropriate with `GrantAppDeployment` / `GrantProvCluster`.
- Adding `app` to the default admin ACL is consistent with exposing this to the app deployment flow.
- The per-cluster pool-info cache is simple and reasonable for the UI/API use case.
- Provisioning still forces a fresh OpenSVC pool lookup, which is the safer behavior before generating the OpenSVC template.
- The cached pool list is cloned before returning, which avoids accidental mutation of cached state.

The OpenSVC V2/V3 pool discovery implementation also looks reasonable. I checked the OpenSVC b2.1 pools handler, and the V2 response shape is driven directly by `pool_status_data(...)`, so the parser approach here seems acceptable for the expected `/get_pools` output. The V3 path using `items` also looks straightforward.

The provisioning changes are also good. `OpenSVCGetAppVolumeSections` now validates that selected pools exist in the runtime OpenSVC pool list and derives the `shared` flag from runtime pool metadata instead of static config. That is a better source of truth. Returning errors up the call chain from template generation is also the right behavior.

The config cleanup is consistent with the new behavior. Removing `ProvAppVolumePools`, `VolumePool`, `PoolTypeLocal`, `PoolTypeRemote`, and `GetAppVolumePools` makes sense now that runtime discovery is the source of truth. This is a behavior/config migration, but it appears intentional for this PR.

Frontend/Redux changes look fine:
- The storage page fetches OpenSVC pools only for OpenSVC orchestrator clusters.
- Volume options now come from runtime pool data.
- Failover topology filters to shared pools, which matches the expected semantics.
- The storage component still supports both string and object pool option shapes, so it is reasonably tolerant.
- Passing `appConfig` down to storage for topology-aware filtering is clean enough.

Tests are a strong part of this PR:
- V2 pool discovery and parsing are covered.
- V3 pool discovery and parsing are covered.
- Shared capability detection is covered.
- Non-2xx API responses are covered.
- ACL coverage for `/opensvc-pools` is added.

Only non-blocking suggestion: the UI could show a clearer empty state when there are no pool options. For example:
- OpenSVC returns no pools
- the pool API fails or ACL blocks it
- failover topology filters out all non-shared pools

Right now the user may just see an empty dropdown. A small message like this would make the behavior easier to understand:

> No OpenSVC storage pools available. Check OpenSVC connectivity and pool configuration.

And for failover:

> No shared OpenSVC pools available for failover topology.

I do not think that should block this PR, though. It can be handled as a follow-up.

From my side: approved.

PR #1485, opened 2026-05-21
feat(billing): add billing profile management

caffeinated92:
## Summary
- add authenticated billing profile GET/PUT proxy endpoints to both API routers
- add billing profile form, state, and service calls in the React billing page
- preserve billing subscription payment and pending-plan compatibility behavior

## Verification
- reviewed staged diff
- git diff --cached --check


caffeinated92:
@claude review

PR #1511, opened 2026-06-11
fix(haproxy): keep service_read fallback in sync with reader state

caffeinated92:
## Summary
- keep HAProxy read fallback decisions in sync with same-pass backend state changes
- defer master read-backend drain/ready decisions until reader availability has been fully evaluated
- add regression tests for same-pass maintenance, recovery, and stale-MAINT transitions

## Problem
In a 2-node cluster, putting the only slave into maintenance could leave HAProxy's `service_read` backend with zero UP servers. The master stayed drained by default while the slave moved to maintenance, and HAProxy did not immediately promote the master back into the read pool.

## Fix
- track the effective read-backend status for entries modified during the current `Refresh()` pass
- compute whether the master should remain in the read backend after processing current reader transitions
- update master read-backend handling so fallback to the master and re-drain of the master both happen in the same refresh pass when needed

## Tests
- added unit coverage for `HasAvailableReader()` and `masterShouldBeReader()`
- added same-pass regression tests for:
  - slave entering maintenance -> master becomes reader
  - slave becoming available again -> master is drained again
  - stale master `MAINT` read entry -> reconciled then drained again when a slave reader exists

### DIFF
--- a/cluster/prx_haproxy.go
+++ b/cluster/prx_haproxy.go
@@ -528,12 +528,18 @@ func (proxy *HaproxyProxy) setLastReadBackendStatus(status string) {
 }
 
 // HasAvailableReader returns true if the read backend currently has at least
-// one entry whose repman-reported state is not stateMaster (i.e. not the
-// current master/leader) and whose effective HAProxy status for this pass is
-// "UP".
+// one entry that is not the current master/leader's own row and whose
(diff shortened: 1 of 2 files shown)


caffeinated92:
@claude review [5f4b35f](https://github.com/signal18/replication-manager/pull/1511/commits/5f4b35f3c66f6e0141b105d6d50b373b350f7b11)

PR #1533, opened 2026-06-20
feat(restic): add repository copy and wipe workflows

caffeinated92:
## Summary
- add restic repository copy support, including saved source presets and copy progress tracking
- add repository wipe workflow updates and relax wipe fetch blocking
- extend restic settings and API support for S3 config mode selection and related UI polish

## Testing
- tested before commit

### DIFF
--- a/cluster/cluster.go
+++ b/cluster/cluster.go
@@ -299,6 +299,7 @@ type Cluster struct {
 	failLoadP12Cert                     bool                        `json:"-"`
 	Mailer                              *mailer.Mailer              `json:"-"`
 	ResticManager                       *backupmgr.ResticManager    `json:"-"`
+	resolvedS3Mode                      string                      // probe-resolved S3 mode during auto startup; "" if not yet probed
 	MessageChan                         chan sharedlog.Message      `json:"-"`
 	ErrorConfigs                        config.ErrorConfigs         `json:"-"` //To store error config
 	Partner                             *config.Partner             `json:"partner" groups:"web"`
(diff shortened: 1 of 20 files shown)


caffeinated92:
Findings
1. Silent copy drop during wipe — cluster/cluster_bck.go:818 · HIGH
  AddCopyTask has no return value. When wipeActive==true, appendTask silently discards the
  task. ResticCopyRepoWithOptions sees no error and returns nil. The HTTP handler responds
  200 "Restic repository copy queued". The copy never runs.

  2. AWS credentials wiped on config reload error — cluster/cluster_bck.go:624 · HIGH
  reloadResticEnvCore initializes all six AWS fields to empty strings, then skips the
  assignment block when resolveResticS3 returns an error. SetAwsConfig is called
  unconditionally with those empty strings, overwriting the live credentials. The API caller
  gets 200 OK; the next backup fails with an S3 auth error.

  3. TaskErrors map mutated without errorMutex — utils/backupmgr/restic.go:3239, 3429, 4405
  · HIGH
  Three bare delete(repo.TaskErrors, InitTask) calls with no lock held. Every other accessor
  (SetError, FetchAndClearErrorsExcept, GetInitIssue) acquires errorMutex first. Concurrent
  access with the monitor loop is a Go data race — detectable with -race, potentially fatal
  in production.

  ---
  4. resolvedS3Mode data race (plausible) — cluster/cluster_bck.go:643, 914 · MEDIUM
  Plain string field written by the boot-fetch goroutine and read by the monitor loop with
  no mutex. Narrow window but a real unsynchronized read/write under the Go memory model.

  5. Dropped fetch tasks on failed wipe (plausible) — utils/backupmgr/restic.go:4189 · LOW
  If the wipe itself fails after draining pending fetch tasks, those tasks are permanently
  discarded with no re-enqueue. Callers are notified with an error, but the metadata refresh
  is lost for that cycle. The monitor re-enqueues on the next tick, so impact is limited to
  one stale cycle.

caffeinated92:
@claude review [720d5c3](https://github.com/signal18/replication-manager/pull/1533/commits/720d5c35c6d0dafe98a00f1cb9b5acfe2ba6c50e)

PR #1550, opened 2026-06-29
fix(app): resolve template placeholders in app env variables on save

caffeinated92:
Add ResolveEnvVariableValue helper that expands {{...}} placeholders in app deployment env variables when they are saved. Unresolvable or partial templates fall back to the raw value so existing behavior is preserved.

Apply the helper to:
- adding a new env variable
- editing an env variable value
- editing env variable conditional values

Secrets are intentionally left unchanged and still go through encryption.

Includes focused tests for full/partial resolution, missing substitution data, conditional values, and secret non-expansion.


caffeinated92:
@claude review PR under 20 steps

PR #1567, opened 2026-07-02
Quiet go-mysql binlog syncer logs and contain constructor failures per cluster

caffeinated92:
This PR hardens replication-managers go-mysql binlog syncer integration in two ways:

- routes `replication.BinlogSyncer` logging through `s18log.BinlogSyncerLogger` under `ConstLogModPurge`, downgrading noisy lifecycle Info logs to DEBUG while preserving Warn/Error visibility
- validates computed binlog syncer server IDs before construction and reports invalid `check-binlog-server-id` values once at cluster startup
- contains `go-mysql` constructor failures per cluster by converting the loggers `Fatal*` path into a recoverable failure, wrapping `replication.NewBinlogSyncer(...)` in a safe constructor, and loudly logging recovered panics instead of exiting the whole process
- keeps PITR/binlog replay safe by propagating `GetBinlogPositionFromTimestamp` failures instead of continuing with an unset stop position
- adds regression tests for the logger adapter, safe constructor recovery, and enforcement that cluster code only constructs binlog syncers through the wrapper

This ensures one clusters bad binlog syncer path cannot take down replication-manager or other monitored clusters while still leaving strong visibility for binlog-related failures.

### DIFF
--- a/cluster/cluster.go
+++ b/cluster/cluster.go
@@ -566,6 +566,27 @@ func (cluster *Cluster) InitFromConf() {
 		cluster.LogModulePrintf(cluster.Conf.Verbose, config.ConstLogModGeneral, config.LvlInfo, "Failover in automatic mode")
 	}
 
+	// go-mysql's replication.NewBinlogSyncer calls Logger.Fatal (os.Exit(1))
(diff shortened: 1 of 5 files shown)


caffeinated92:
@claude review

PR #1574, opened 2026-07-10
fix(inject): dml traffic marker via proxy create-once (no replication break)

svaroqui:
## What

Correct the `inject-traffic-mode=dml` marker so it no longer breaks replication.

## Why

The DML marker created the pseudo-GTID table **out-of-band on each node** (`ensureInjectTrafficTable`, currently on develop). The master's per-tick in-place `REPLACE` of row `id=1` then replicated as a ROW update whose *before-image* the empty/independently-created slave table didn't have → `1032 HA_ERR_KEY_NOT_FOUND`, slave goes `SlaveErr`. Observed live on belair/db1.

This is why the original marker was a `CREATE OR REPLACE VIEW`: self-contained DDL that needs no table on newly-monitored/reseeded nodes.

## Fix

Create the table the **same way the marker is written — via the proxy, once, when traffic starts**. That CREATE lands on the master, is binlogged, and flows to every slave in the **same binlog stream before** any `REPLACE` that follows it. Binlog ordering guarantees a slave applies the CREATE before the REPLACE (no 1032), and a reseeded node gets the table in its snapshot. So there is **nothing to check** — no `CanInjectTraffic` state, no per-node presence probe, no `information_schema`/`SHOW CREATE` lookup, no new-node tracking. One idempotent create-once guard.

- Default is `ddl` (self-contained view, battle-tested); `dml` is experimental/opt-in and topology-aware (`force-slave-no-gtid-mode` keeps the greppable view for positional rejoin).
- `injectTrafficMarker` creates DB+table+seed+view via the proxy `db` on first call, then `REPLACE`s — all on the same connection.
- Removes the earlier over-engineered state machine.

## Testing note

Touches the traffic-injection / replication path — must run through the OpenSVC **topology matrix** (master-slave GTID + master-master old-style positional, across db/proxy versions) before merge. Do not merge until the matrix is green.

### DIFF
--- a/cluster/cluster.go
+++ b/cluster/cluster.go
@@ -96,7 +96,7 @@ type Cluster struct {
 	Status                        string                 `json:"activePassiveStatus" groups:"web"`
 	IsSplitBrain                  bool                   `json:"isSplitBrain" groups:"web"`
 	IsSplitBrainBck               bool                   `json:"-"`
-	injectTrafficTableReady       bool                   `json:"-"` // dml marker schema created once via the proxy
+	injectTrafficTableReady       map[string]bool        `json:"-"` // dml marker schema created once per proxy target
 	IsFailedArbitrator            bool                   `json:"isFailedArbitrator" groups:"web"`
 	IsLostMajority                bool                   `json:"isLostMajority" groups:"web"`
 	IsDown                        bool                   `json:"isDown" groups:"web"`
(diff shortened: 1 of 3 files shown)


svaroqui:
@claude please review this PR — focus on the inject-traffic DML marker correctness (binlog ordering guarantee that the create-once-via-proxy CREATE reaches slaves before any REPLACE) and any topology cases it could break.

PR #1593, opened 2026-07-17
fix(startup): guard nil CheckSumConfig — boot panic when subscription-plan sync saves during InitConfig

svaroqui:
## Crash

```
server.(*ReplicationManager).SaveDynamic          server/server.go:4222
server.(*ReplicationManager).SaveGlobalConfigs    server/server.go:4311
server.(*ReplicationManager).SaveCallBack         server/server.go:4294
config/manager.(*ConfigManager).SaveConfig        config/manager/manager.go:388
server.(*ReplicationManager).persistInstanceSubscriptionPlan   server/api_register.go:947
server.(*ReplicationManager).syncSubscriptionPlanFromCRM(WithToken)  server/api_register.go:1036/1050
server.(*ReplicationManager).InitConfig           server/server.go:1955
```

## Root cause

`syncSubscriptionPlanFromCRM()` runs at the **tail of `InitConfig`** (server.go:1955) and persists the plan **synchronously** — `SaveConfig(repman, true)` → `SaveGlobalConfigs` → `SaveDynamic`. But `CheckSumConfig` isn't created until **line 2374**, a later startup step. `SaveDynamic` writes `repman.CheckSumConfig["saved"] = new_h` to a **nil map → panic ("assignment to entry in nil map")**, crashing on boot.

Surfaced by the #1591 merge: the config-sync rework (`SaveConfig`/`SaveGlobalConfigs`) met develop's subscription-plan sync, which fires a save earlier in startup than any save ran before.

## Fix

Lazily initialize `CheckSumConfig` at the top of `SaveGlobalConfigs` — order-independent, so any early save finds the map ready. Config **content is unaffected**: `Conf`, `DefaultFlagMap` (server.go:299) and `ImmuableFlagMap` (server.go:1910) are all populated before the sync runs, so the persisted `default.toml` is correct.

## Validation

- `go build --tags server` passes (pro binary).
- Fix is on the exact panic path and covers `SaveDynamic` + `SaveImmutable` reached via `SaveGlobalConfigs`.

Blocks the **v3.1.32** release — recommend merge + rebuild.

### DIFF
--- a/server/config_save_test.go
+++ b/server/config_save_test.go
@@ -1,0 +1,30 @@
+package server
+
+import (
+	"testing"
+
+	"github.com/signal18/replication-manager/config"
+)
+
+// TestSaveGlobalConfigs_NilCheckSumConfig pins the boot-crash regression fixed by
+// the nil-map guard in SaveGlobalConfigs. A config save can be triggered before
+// CheckSumConfig is allocated — syncSubscriptionPlanFromCRM persists the plan
+// synchronously at the tail of InitConfig, well before the map is created later in
+// startup. Writing to a nil map panics; the save path must instead find it ready.
+func TestSaveGlobalConfigs_NilCheckSumConfig(t *testing.T) {
+	repman := &ReplicationManager{
+		Conf: &config.Config{
+			ConfRewrite: true,
+			WorkingDir:  t.TempDir(),
+		},
+	}
+
+	// CheckSumConfig left nil on purpose, reproducing the early-startup state.
+	if err := repman.SaveGlobalConfigs(); err != nil {
+		t.Fatalf("SaveGlobalConfigs returned an error with nil CheckSumConfig: %v", err)
+	}
+
+	if repman.CheckSumConfig == nil {
+		t.Fatal("SaveGlobalConfigs must lazily initialize CheckSumConfig when nil")
+	}
+}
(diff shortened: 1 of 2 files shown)


svaroqui:
@claude please re-review. Since your last pass I addressed the review feedback:

- **Regression test added** (`server/config_save_test.go`): instantiates `ReplicationManager` with `CheckSumConfig == nil` and asserts `SaveGlobalConfigs` neither panics nor errors and initializes the map. Passes under `go test --tags server`.
- **Follow-up filed (#1594)** for the durable source-fix: allocate `CheckSumConfig` once at construction so direct writers (`ReloadTerms`, `Overwrite`, `SaveDynamic`, `SaveImmutable`) don't depend on the `SaveGlobalConfigs` guard.
- **Concurrency note** (unsynchronized `CheckSumConfig` writes) recorded as pre-existing/out-of-scope in #1594.

The guard here stays as the minimal, order-independent hotfix for the release-blocking boot panic. Please confirm it's sound and flag anything else.

svaroqui:
Folded in the durable source fix (was #1594, now closed): `CheckSumConfig` is allocated at construction in `SetDefaultFlags`, so all direct writers are nil-safe — the guard here is now just insurance. Added tests for the source init and a direct `SaveDynamic` call. `go build --tags server` and the three `config_save_test.go` cases pass.

PR #1595, opened 2026-07-17
fix(startup): allocate CheckSumConfig at construction (durable follow-up to #1593)

### LINKED ISSUE #1594: Init ReplicationManager.CheckSumConfig once at construction (durable fix behind #1593 guard)
## Context

The boot crash fixed in #1593 was a write to a nil `ReplicationManager.CheckSumConfig` map: `syncSubscriptionPlanFromCRM()` persists synchronously at the tail of `InitConfig` (server.go:1955), but `CheckSumConfig` isn't allocated until later in startup (server.go:2374). #1593 guards it with a lazy `if repman.CheckSumConfig == nil { make(...) }` at the top of `SaveGlobalConfigs`.

## Why this is a follow-up, not the end of it

The guard sits in **one** function. `CheckSumConfig` has **other direct writers** that bypass it:

- `ReloadTerms()` → `CheckSumConfig["terms"]` (server.go:3941)
- `Overwrite()` → `CheckSumConfig["overwrite"]` (server.go:4135)
- `SaveDynamic()` / `SaveImmutable()` write it directly too — safe **only** because they're currently reached via `SaveGlobalConfigs`.

Today those are all invoked after the map exists, so there's no live bug beyond the one #1593 fixed. But the pattern is fragile: any **new early caller** of a direct writer (as `ReloadTerms` already is) would reintroduce the same class of panic, and the `SaveGlobalConfigs` guard wouldn't cover it. (Raised in code review of #1593.)

## Proposed durable fix

Allocate `CheckSumConfig` **once, unconditionally, at the struct's startup initialization** — alongside the other early maps (e.g. next to `repman.DefaultFlagMap = make(...)` in `SetDefaultFlags`, server.go:299), so it is never nil for the life of the instance. Then:

- the per-`InitConfig` `make` at server.go:2374 becomes a deliberate "reset checksums on reload" rather than the first allocation, and
- the lazy guard in `SaveGlobalConfigs` can be removed (or kept as cheap insurance).

## Out of scope (noted, not this ticket)

`CheckSumConfig` writes aren't mutex-synchronized. Pre-existing, unrelated to #1593 — flag separately if concurrent saves become a concern.

## Done when
- [ ] `CheckSumConfig` allocated at construction/early init; never nil.
- [ ] Direct writers (`ReloadTerms`, `Overwrite`, `SaveDynamic`, `SaveImmutable`) safe without relying on the `SaveGlobalConfigs` guard.
- [ ] Regression test extended to a direct writer (e.g. `SaveDynamic` on a nil-map instance).

svaroqui:
Follow-up to #1593 (whose guard is already on develop and fixes the boot crash). This adds the **durable source fix** that missed the #1593 merge by one commit.

## Change
Allocate `CheckSumConfig` in `SetDefaultFlags` — the earliest startup init, before `InitConfig` and any save — so **every** direct writer (`SaveDynamic`, `SaveImmutable`, `Overwrite`, `ReloadTerms`) is nil-safe, not just the `SaveGlobalConfigs` path. Addresses the review point that the `SaveGlobalConfigs`-local guard wouldn't cover future early direct writers (as `ReloadTerms` already is).

The `SaveGlobalConfigs` guard stays as cheap defensive insurance for callers that bypass `SetDefaultFlags` (e.g. tests). Conditional (`if nil`) so the repeated `SetDefaultFlags` calls from secret-store setup don't reset live checksums.

## Tests (all pass under `go test --tags server`)
- `SetDefaultFlags` allocates the map.
- A direct `SaveDynamic` after `SetDefaultFlags` doesn't panic (exercises a writer with no local guard).
- The `SaveGlobalConfigs` guard still covers a bare nil-map instance.

`go build --tags server` (pro) passes. Closes #1594.

### DIFF
--- a/server/config_save_test.go
+++ b/server/config_save_test.go
@@ -4,13 +4,43 @@ import (
 	"testing"
 
 	"github.com/signal18/replication-manager/config"
+	"github.com/spf13/viper"
 )
 
-// TestSaveGlobalConfigs_NilCheckSumConfig pins the boot-crash regression fixed by
-// the nil-map guard in SaveGlobalConfigs. A config save can be triggered before
-// CheckSumConfig is allocated — syncSubscriptionPlanFromCRM persists the plan
-// synchronously at the tail of InitConfig, well before the map is created later in
-// startup. Writing to a nil map panics; the save path must instead find it ready.
+// TestSetDefaultFlags_InitsCheckSumConfig verifies the source fix: CheckSumConfig is
+// allocated at the earliest startup init (before InitConfig and any save), so no
+// config-save path can hit a nil map.
+func TestSetDefaultFlags_InitsCheckSumConfig(t *testing.T) {
+	repman := &ReplicationManager{Conf: &config.Config{}}
+
+	repman.SetDefaultFlags(viper.New())
+
+	if repman.CheckSumConfig == nil {
+		t.Fatal("SetDefaultFlags must allocate CheckSumConfig")
+	}
+}
+
+// TestSaveDynamic_NilSafeAfterSetDefaultFlags exercises a direct writer. SaveDynamic
+// writes CheckSumConfig itself and is not individually guarded — the SetDefaultFlags
+// init is what keeps it (and ReloadTerms, Overwrite) nil-safe. Pins that guarantee.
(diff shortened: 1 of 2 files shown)


caffeinated92:
@claude review

PR #1600, opened 2026-07-17
feat(schema): add schema advisor plugins

### LINKED ISSUE #1592: Schema advisor plugins: wide-varchar row-size check + uncompressed large BLOB/TEXT check (MariaDB)
## Summary

Add **schema advisor plugins** (schema-check rules) that flag row-layout / storage anti-patterns that hurt InnoDB performance and can even make rows unstorable. Two first rules:

### Rule 1 — Wide variable rows: `SUM(varchar) > InnoDB page size`

Flag tables where the combined size of the `VARCHAR` columns pushes the row past what InnoDB can store inline. InnoDB has a hard **row-size limit (~8126 bytes on a 16K page, half-page)**; wide variable rows either fail with *"Row size too large"* or force off-page storage that kills scan performance.

- Check: sum of `VARCHAR` column widths (focus on the short, inline-stored ones — declared length `< 256`, which InnoDB stores inline with a 1-byte length prefix) exceeds the page-size / row-size budget.
- Report: table name, offending columns, computed sum vs. the InnoDB limit, and the effective `innodb_page_size` / row format.
- Threshold: `< 256` bytes is the inline/off-page boundary; the row-size limit (page-size vs. half-page) to be confirmed during implementation.

### Rule 2 — Uncompressed large BLOB/TEXT on MariaDB

Flag tables that have `BLOB`/`TEXT` columns with **average size > ~8K** that are **not** using MariaDB **per-column compression** (`column_name BLOB/TEXT ... COMPRESSED`).

- Rationale: large uncompressed LOB columns waste storage and I/O; MariaDB's per-column compression is a cheap win for exactly this case.
- Check: for each `BLOB`/`TEXT` column, compare `AVG(LENGTH(col))` against the threshold (~8K) and detect whether the column declares `COMPRESSED`.
- Report: table + column, average size, and a suggestion to enable per-column compression.
- MariaDB-only (per-column `COMPRESSED` is a MariaDB feature).

## Notes
- These live as schema-check / advisor rules surfaced in the GUI schema view.
- Thresholds (`< 256`, ~8K, page-size vs. half-page) are starting points — confirm against the InnoDB row-size math and real MariaDB behavior when wiring the checks.

caffeinated92:
## Summary
- add two schema advisor plugins for issue #1592: wide inline VARCHAR row risk and large uncompressed MariaDB BLOB/TEXT columns
- preserve external plugin SCHEMA severity and route schema findings through the existing schema state path
- extend the schema wire snapshot with row-size and LOB metadata used by schema advisors
- enrich MariaDB schema scans with bounded-sample LOB sizing and per-column COMPRESSED detection
- add unit tests for the new plugins, schema wire behavior, and LOB enrichment helpers

## Details
This PR implements the first schema advisor rules on top of the existing schema dictionary snapshot produced by schema monitoring.

### Rule 1: wide variable rows
Flags InnoDB tables whose short inline-stored VARCHAR columns collectively exceed the estimated inline row budget.

### Rule 2: large uncompressed LOBs
Flags MariaDB BLOB/TEXT columns whose observed average size is above the configured threshold and that are not using per-column COMPRESSED storage.

## Files of note
- cluster/logplugin/external.go
- cluster/srv_log_plugins.go
- cluster/logplugin/plugins/plugin-schema-row-size/*
- cluster/logplugin/plugins/plugin-schema-lob-compression/*
- utils/dbhelper/schema_lob.go
- cluster/logplugin/plugins/wire/wire.go

## Notes
- schema advisor results are surfaced through the existing schema domain / schema view path
- LOB sizing uses bounded sampling instead of full-table aggregation to avoid expensive scans on large tables

Closes #1592

### DIFF
--- a/cluster/cluster.go
+++ b/cluster/cluster.go
@@ -2708,6 +2708,17 @@ func (cluster *Cluster) MonitorMasterTableSchema() error {
 
 	cluster.WorkLoad.DBIndexSize = totindexsize
 	cluster.WorkLoad.DBTableSize = tottablesize
+
+	// Enrich BLOB/TEXT candidate columns with MariaDB compression detection and
+	// a bounded-sample observed average byte length, consumed by the
+	// plugin-schema-lob-compression schema advisor plugin. No-op on non-MariaDB
+	// and for tables under the minimum candidate size; every query is bounded
+	// and failures are skipped silently — see dbhelper.EnrichLobColumns.
+	if cluster.Conf.MonitorSchemaColumns {
+		lobLogs := dbhelper.EnrichLobColumns(cmaster.Conn, cmaster.DBVersion, tables)
+		cluster.LogSQL(lobLogs, nil, cmaster.URL, "Monitor", config.LvlDbg, "LOB column enrichment on %s", cmaster.URL)
+	}
+
 	cmaster.DictTables = dbhelper.FromNormalTablesMap(cmaster.DictTables, tables)
 
 	return nil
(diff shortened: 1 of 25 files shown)


caffeinated92:
@claude review

PR #1606, opened 2026-07-18
fix(peer): health-poll scope + socket-leak, SSO-login root cause, marketplace/for-sale GUI

svaroqui:
## Summary

Fixes the peer-health / marketplace polling so a repman never adds load to
partners it has no relationship to, and so legitimate fleet polling can't strand
connections on a dark peer. Two independent problems, two fixes, plus a new
technical doc.

Production symptom this addresses: one repman went unresponsive with ~14k
goroutines and ~6.9k TCP connections wedged against a single peer.

## Fix 1 — scope: no traffic on partners from registrants

A fresh/browsing repman now opens **zero** seller connections. All live polling
routes through the connected-users scope (`GetHealthStatusForActiveUsers`):

- `BatchUpdateClusters` no longer polls (it has no session-user context, so it
  could only walk the flat, unscoped list — for-sale included).
- The `peer.json` reload path calls `dispatchPeerHealthPoll()` after updating
  data, so refresh stays instant-on-pull **and** scoped.
- `pulling` and `smart` both route to the scoped poller; the unscoped
  `GetHealthStatusForUnknownVersions` is removed.
- `peering` remains the single opt-in unscoped full-mesh (never the default).

A for-sale cluster is polled only once a connected user enters its **sale
workflow** (`pending`/`sponsor` role) or is delegated it — and then only that one
seller, never the rest of the catalog. Catalog status comes from the
BO-aggregated `peer.json` (git-pull cadence), which is what keeps the sale view
fast.

## Fix 2 — leak: legitimate fleet polling must not strand connections

Fleet polling (your own repmans + any repman sharing a cluster with you or a
connected SSO user) *should* generate traffic, so the leak is fixed for that path:

- **Bounded transport** — `MaxConnsPerHost: 2` hard ceiling (thousands of sockets
  to one peer is now impossible), plus dial/handshake/response deadlines for a
  peer that accepts TCP but never answers.
- **Rate-limit + advance `LastUpdate` on failure** — each peer's login+health is
  gated to once per `Interval`, advanced up-front so a failed attempt is
  rate-limited like a success. (The core bug: failures never advanced the gate and
  login wasn't gated at all, so a dark peer was re-hit every dispatch.)
- **Effective single-flight** — `dispatchPeerHealthPoll` self-guards on
  `peerHealthBusy` and holds it across the whole poll, so timer + reload can't
  overlap and race the shared `LastUpdate`.

## Docs

New `doc/implementation/peer/MARKETPLACE.md`: the delegation-vs-for-sale model,
the fleet/identity rules, the invariant (with the sale-workflow and
always-on-vs-session precision), the sale-workflow timing, and both fixes.

## Known follow-up (not in this PR)

- **Identity `admin` → `Cloud18GitUser` (§2.1).** Own clusters that carry only the
  local `admin` ACL entry may not land under `PeerUserClusters[Cloud18GitUser]`,
  so 24/7 own-fleet polling for master-failed alerting isn't yet guaranteed for
  them. This is a BO/`peer.json` change.
- Alerting design: whether "in charge" can be a sub-user, and whether the alert is
  raised locally from peer `IsMasterDown` or by the hosting repman.

## Testing

`go build ./peer/` and `go build -tags server ./server/` pass. Behavioural
verification (fresh instance = zero seller connections; fleet polling survives a
dark peer) to be exercised on the OpenSVC test cluster.

### DIFF
--- a/doc/implementation/peer/MARKETPLACE.md
+++ b/doc/implementation/peer/MARKETPLACE.md
@@ -335,13 +335,21 @@ that traffic — a single dark peer in your own fleet must not accumulate connec
    success (the claim is advanced up-front, before the call). Previously a failure
    left `LastUpdate` untouched and login wasn't even gated, so a dark peer was re-hit
    every dispatch.
-3. **Correct map locking, no single-flight** (`peer.go`) — because the claim, the
-   `PeerURL`/`Clients` access, and `GetHealthStatus`'s `PeerClusters` update all hold
-   `pm.mu` (released before network I/O), overlapping timer + reload dispatches can no
-   longer race the maps (`BatchUpdateClusters` mutates them under the same lock). So
-   no whole-poll serialization is needed: **different** peers still poll in parallel;
-   only the **same** peer is deduplicated. The earlier `peerHealthBusy` single-flight
-   guard was dropped in favour of this finer-grained, correct-by-construction claim.
+3. **Correct locking** (`peer.go`) — the claim, the `PeerURL`/`Clients` access, and
+   `GetHealthStatus`'s `PeerClusters` update all hold `pm.mu` (released before network
+   I/O), so a poll goroutine never races `BatchUpdateClusters` (which mutates the same
+   maps under the same lock). `pollPeerHealth` also writes `nodestat.Error` under the
+   lock (via `setNodeError`) because the status API (`GetPeerNodesJSON`) reads those
+   fields — and `GetPeerNodesJSON` now marshals a value-copy **snapshot** taken under
+   the lock, never the live pointers. Dedup is per-peer (different peers still poll in
+   parallel; only the **same** peer is skipped), so no whole-poll serialization is
+   needed for correctness.
+4. **Coarse busy signal, kept for operators** (`server_peer.go`) —
+   `dispatchPeerHealthPoll` still self-guards on `peerHealthBusy` and holds it for the
+   whole poll, purely so the flag reflects "a poll is running." The timer raises
+   `GWARN013@peerhealth` when it finds a poll that hasn't finished within a cycle (a
+   slow/stuck poll — the signal that would have surfaced the original incident). This
+   is complementary to, not a replacement for, the per-peer claim.
 
 **Timing (why Fix 1 costs no delay):** periodic backstop dispatch every ~2 min
 (`counter%60` × `monitoring-ticker` 2s); per-node staleness gate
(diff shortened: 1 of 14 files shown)


svaroqui:
@claude please re-review the latest commit 6352774c8 ("first-in-wins atomic claim"), which landed after the initial automated review. Focus on concurrency correctness in `peer/peer.go`:

1. `pollPeerHealth` — the atomic claim: it takes `pm.mu`, checks `time.Since(nodestat.LastUpdate) > Interval`, advances `LastUpdate`, gets/creates the client, then **releases the lock before** `PeerLogin`/`GetHealthStatus`. Is the claim race-free (can two dispatches both poll the same peer)? Is it correct that `nodestat.Error` and the returned client are used off-lock (single-owner-per-interval assumption)?
2. `GetHealthStatus` — now takes `pm.mu` around the `PeerClusters` mutation, after the network call. Any lock-ordering / re-entrancy issue given the caller already released the lock?
3. `GetAllHealthStatus` — snapshots URLs under RLock then polls off-lock. OK vs iterating the live map?
4. Confirm no path holds `pm.mu` across network I/O, and that all of this stays off the main server loop (called from the timer goroutine and the cloud18-pull goroutine).
5. `dispatchPeerHealthPoll` dropped its `peerHealthBusy` single-flight in favour of the per-peer claim — any regression from overlapping timer+reload dispatches?

Also sanity-check `peer/client.go` transport bounds (`MaxConnsPerHost: 2`, dial/handshake/response deadlines) and that Fix 1 scoping (no for-sale fan-out for a fresh browser) still holds.

svaroqui:
@claude the first-review findings are now addressed in commit 2d6540a5d — please re-review the full PR at its current head (`2d6540a5d`). Specifically confirm:

1. **HealthMode race (finding #1)** — new locked `PeerManager.SetHealthMode()` used at all three write sites (`server.go` InitConfig, `api_register.go`, `api_global_settings.go`, the last with a nil-guard). Is the race fully closed, and are there any other unlocked reads/writes of `HealthMode`?
2. **PeerLogin (finding #4)** — non-OK login now returns an error instead of nil. Correct?
3. **Flag help text (finding #2)** — reworded; accurate now?
4. **Tests (finding #3)** — new `peer/peer_scope_test.go` + extracted `relevantPeerURLs()`. Do the tests actually guard the invariant (for-sale excluded, fresh browser polls nothing, ReloadUsers pending/sponsor demotion)? Any gap?

Also please carry forward the concurrency review of commit 6352774c8 (first-in-wins atomic claim in `pollPeerHealth` / `GetHealthStatus` / `GetAllHealthStatus`) if it was not already covered, and give an overall merge verdict for the branch.

PR #1612, opened 2026-07-20
fix(reload): preserve live cluster state during config reload

caffeinated92:
## Summary

This PR fixes a chain of live cluster config reload issues that showed up on running clusters, especially on deployments where the static cluster config lives in `/etc/replication-manager/cluster.d/*.toml` and dynamic overlays live under the cluster working directory.

Before this branch, a reload could:

- use the wrong flag maps after `InitConfig`, breaking secret decryption behavior
- rebuild global server state for a single-cluster reload
- silently relocate the live working directory on non-root runs
- lose static cluster settings from `default.include` / `cluster.d`
- reload with no `db-servers-hosts` / proxy hosts
- race the active monitor tick while rebuilding cluster state
- block too long while draining in-flight tick work

This PR makes live reload:

- cluster-scoped instead of global
- consistent with startup config layering
- safe for include-based static cluster config
- serialized against the active monitor tick
- bounded when draining tracked tick goroutines

### Global rebuild for a single-cluster reload
The old reload path called `InitConfig(*repman.Conf, true)` and then reloaded one cluster. That re-ran global discovery and mutated:

- `repman.ClusterList`
- `repman.ImmutableClusterList`
- `repman.Confs`
- `repman.VersionConfs`
- per-cluster flag maps

That was too broad for `/api/clusters/{cluster}/settings/actions/reload`.

### Secret decryption regressions after reload
After `InitConfig()`, the reloaded cluster config could still point at the wrong flag maps, which broke secret decryption behavior and led to downstream credential issues.

### Working directory relocation on non-root reloads
Re-running `InitConfig()` during reload could re-derive `WorkingDir` from fallback logic and silently move a live cluster into `~/.local/replication-manager/data`, causing some subsystems to behave like first boot.

### Static cluster config lost on live reload
The first isolated reload implementation only read the saved dynamic overlay from the working directory. It did not reconstruct the static cluster section from the main config/include directory.

On layouts like:

- `/etc/replication-manager/config.toml`
- `include = "/etc/replication-manager/cluster.d"`
- `/etc/replication-manager/cluster.d/dev2.toml`

reload could log:

- `Could not parse configuration group`
- `No hosts list specified`

and then fail to rebuild DB servers and proxies.

### Reload vs monitor tick race
`cluster.ReloadConfig()` rebuilt `Conf`, state machines, `Servers`, `Proxies`, and topology in place while `cluster.Run()` could still be executing the current tick and some of its fire-and-forget goroutines.

### Reload drain could block too long
Once reload/tick serialization was added, reload could wait too long on tracked tick goroutines, making the cluster appear stuck.

## Root cause

The reload path was mixing startup/global config rebuild semantics with live single-cluster reload semantics.

Startup config assembly correctly loads:

- main config
- include dir files
- `[DEFAULT]`
- `[saved-default]`
- per-cluster static sections
- per-cluster saved overlays

But it also rebuilds global runtime state, which is unsafe for a targeted live reload.

Separately, `cluster.ReloadConfig()` assumed it could rebuild cluster state in place without coordinating with the active monitor loop.

#### Preserve cluster flag maps on reload
Added `ReloadClusterConfig()` so reload rebinds:

- `ImmuableFlagMap`
- `DynamicFlagMap`
- `DefaultFlagMap`

before applying the reloaded config.

#### Stop using full `InitConfig()` for cluster-scoped reload
HTTP and gRPC reload paths now use a cluster-only reconstruction path instead of `InitConfig()`.

Updated:

- `server/api_cluster.go`
- `server/repmanv3.go`
- `server/regtest.go`

#### Reconstruct live cluster config in isolation
Added live reload reconstruction in `server/server.go` that rebuilds only the requested cluster’s config from disk/runtime context.

This reconstruction now loads:

- main config via the same search rules as startup
- static cluster files from `default.include`
- `WorkingDir/default.toml` for `[saved-default]`
- `WorkingDir/<cluster>/<cluster>.toml` for `[saved-<cluster>]`

#### Restore startup-equivalent default layering for reload
Reload now reapplies:

- `[DEFAULT]`
- `[saved-default]`

onto the live baseline before merging the target cluster section and saved cluster overlay.

#### Preserve live runtime-derived fields
Reload re-applies derived runtime fields like:

- `Interactive`
- `MonitorAddress`
- `ShareDir`
- `WorkingDir` when based on `BaseDir`

without re-running the unsafe global fallback path.

#### Add reload regression coverage
Added `server/server_reload_test.go` covering:

- include-based static cluster section loading
- config auto-discovery when `--config` is not set
- preservation of live working directory
- no sibling cluster mutation during reload
- pickup of changed `[DEFAULT]` / `[saved-default]`
- logging when include-dir reading fails

#### Serialize reload against the active monitor tick
Added a dedicated per-cluster `reloadMu` in `cluster/cluster.go`.

`cluster.Run()` now executes the synchronous tick body under this mutex, and `ReloadConfig()` takes the same mutex before rebuilding cluster state.

#### Track tick-spawned fire-and-forget goroutines
Added `tickWG` and `trackTickGoroutine(...)` to cover the tick-spawned async work that is not already joined by local `WaitGroup`s.

This now includes relevant operations like:

- `CheckDefaultUser`
- `MonitorSchema`
- `initOrchetratorNodes`
- `ResticFetchRepo`
- `RefreshAllAppTemplateMD5`
- `GetSlowLogTable`
- jobs upgrade / reseed / restic reseed paths
- snapshot metadata reconciliation async path

#### Drain tracked tick work before rebuilding
`ReloadConfig()` now waits for tracked in-flight tick goroutines before rebuilding the cluster.

#### Bound reload drain wait
Added a timeout-based drain helper so reload does not wait forever on long operational jobs. If the drain exceeds the bound, reload proceeds with a warning instead of silently freezing the cluster monitor loop indefinitely.

## User-visible behavior after this PR

Reloading a cluster should now:

- keep the cluster present in the running inventory
- keep static cluster settings from `/etc/replication-manager/cluster.d/*.toml`
- keep DB host lists and proxy host lists intact
- preserve secret decryption behavior
- avoid unexpected working-dir relocation on non-root installs
- avoid racing the active monitor tick
- avoid waiting forever on tracked long-running tick work

## Validation

Passed:

- `go test ./server/...`
- `go test ./cluster ./server`

## Files changed

- `server/api_cluster.go`
- `server/regtest.go`
- `server/repmanv3.go`
- `server/server.go`
- `server/server_reload_test.go`
- `cluster/cluster.go`
- `cluster/cluster_bck_meta.go`

## Commits in this branch

- `fix(server): preserve cluster flag maps on reload`
- `fix(server): isolate live cluster config reload`
- `fix(server): restore static/default config on live reload`
- `fix(cluster): serialize live reload with monitor tick`
- `fix(cluster): bound live reload drain wait`

### DIFF
--- a/cluster/cluster.go
+++ b/cluster/cluster.go
@@ -945,6 +945,31 @@ func (cluster *Cluster) Run() {
 	}
 }
 
+// reloadDrainTimeout bounds how long ReloadConfig waits for tickWG to drain.
+// See waitTickDrain and its call site in ReloadConfig for why this must not
+// be indefinite.
+const reloadDrainTimeout = 10 * time.Second
+
+// waitTickDrain waits for tickWG to reach zero, up to timeout. Returns true
+// if it drained fully, false if the timeout elapsed first. The Wait() runs in
+// its own goroutine so a timeout can be observed without blocking the
+// caller; if tickWG never drains, that goroutine leaks for the life of the
+// process, which is an accepted tradeoff for a codepath that should be rare
+// (see ReloadConfig's caller).
+func (cluster *Cluster) waitTickDrain(timeout time.Duration) bool {
+	done := make(chan struct{})
+	go func() {
+		cluster.tickWG.Wait()
+		close(done)
+	}()
+	select {
+	case <-done:
+		return true
+	case <-time.After(timeout):
+		return false
+	}
+}
+
 // trackTickGoroutine launches fn in a goroutine tracked by tickWG. Use it for
 // a tick's genuinely fire-and-forget spawns -- ones not already joined by a
 // local sync.WaitGroup within the same tick (e.g. the wg/goRun-based Phase
@@ -2170,7 +2195,21 @@ func (cluster *Cluster) ReloadConfig(conf config.Config) {
 	// the tick that just released reloadMu (see tickWG's declaration). Safe
 	// to call here: reloadMu being held blocks Run() from starting a new
 	// tick, so no concurrent Add() can race this Wait().
-	cluster.tickWG.Wait()
+	//
+	// Bounded, not indefinite: some tracked work (rolling jobs upgrade,
+	// script/reseed deployment, restic fetch) is a genuinely long per-server
+	// operational job that can run for minutes, not the quick cluster-state
+	// touch tickWG exists to catch. Waiting on it forever here would freeze
+	// the whole monitor loop for this cluster -- reloadMu stays held for as
+	// long as the wait does, and Run() needs reloadMu for every subsequent
+	// tick, so a stuck drain silently stops the ticker with no panic (the
+	// stuck goroutine itself is doing real work, not asleep). Past the
+	// timeout, proceed anyway: residual overlap with one long job is a
+	// smaller risk than reload never completing at all.
+	if !cluster.waitTickDrain(reloadDrainTimeout) {
+		cluster.LogModulePrintf(true, config.ConstLogModGeneral, config.LvlWarn,
+			"ReloadConfig(%s): proceeding after %s with tick goroutines still running", cluster.Name, reloadDrainTimeout)
+	}
 
 	cluster.Lock()
 	*cluster.Conf = conf
--- a/cluster/cluster.go
+++ b/cluster/cluster.go
@@ -78,107 +78,107 @@ type ClusterResponse struct {
 }
 
 type Cluster struct {
-	OsUser                        *user.User             `json:"-"`
-	Name                          string                 `json:"name" groups:"apps,web"`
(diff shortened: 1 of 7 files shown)


caffeinated92:
@claude review [b5ba58f](https://github.com/signal18/replication-manager/pull/1612/commits/b5ba58f2d533954a071dcfe5e7612bf374a93f20)

svaroqui:
@claude please review this PR. Focus areas: correctness of preserving live cluster state across a config reload (no loss of in-memory monitoring/topology state), the reload path (`InitConfig` re-read + `ReloadConfig`), any races between the reload and the monitoring loop, and regressions to per-cluster saved-overlay vs seed `cluster.d` merge behavior.

PR #1613, opened 2026-07-20
fix(crash-pill): stop blinking on a healthy cluster with stale rejoin history

svaroqui:
## Problem

A **healthy** cluster with old crash history kept the crash pill **blinking red**. Proven on the **belair** test cluster (s18): `db1` crashed twice (both `not-flashback-able`), then became a healthy slave of the elected master `db2`.

Two layers disagreed:
- **GUI** blinked on `failoverHistory.some(c => REJOIN_FAILED.includes(c.rejoinResult))` — *any* record over *all* history.
- **Backend** reconciler `assertRejoinResultStates` rewrote only the **latest** failed record per URL to `recovered`, and required the node be a healthy slave of the crash's historical `ElectedMasterURL`.

So the older `not-flashback-able` record was orphaned → `.some()` stayed true → the pill blinked forever on a fully recovered cluster. (Second latent trigger: after a later switchover, current master ≠ `ElectedMasterURL` → never reconciled.)

## Fix

**Backend (`cluster/srv_lostevents.go`)**
- New `isRejoinResolved()`: a node is resolved when it is the **current** master or an active slave of a **live** master — keyed on the current topology, not the crash's historical `ElectedMasterURL` (so a later legitimate switchover no longer keeps an old failure alarming).
- `assertRejoinResultStates()` now reconciles **every** failed record for a resolved server to `recovered`, not just the latest; still raises the operator warn once per still-unresolved server on its latest failure.

**Frontend (`utility/crashPill.js`, Navbar, HADetail)**
- `crashRejoinBlinks()`: blink/red reflect the **latest rejoin outcome per server**, not `.some()` over all history, so a stale failure superseded by a recovery no longer blinks.

## Tests
- Go: `TestIsRejoinResolved`, `TestReconcileAllFailedRecords` (belair repro — two `not-flashback-able` → both recover), `TestReconcileLeavesUnresolvedAlone`.
- JS: `crashRejoinBlinks.test.js` (11 assertions; proves the old `.some()` logic blinked and the new one does not).

## Verified live
Deployed to the s18 active test node; belair's orphaned `not-flashback-able` reconciled to `recovered` — pill no longer blinks.

Orchestrator/rejoin code — validated on the OpenSVC topology-matrix test cluster.

### DIFF
--- a/doc/implementation/ui-components/DASHBOARD_HEALTH_SIGNALS.md
+++ b/doc/implementation/ui-components/DASHBOARD_HEALTH_SIGNALS.md
@@ -1,0 +1,77 @@
+# Dashboard Health Signals — state machine, not archives
+
+A dashboard health light (pill / badge / alert dot) reflects the cluster's **current health**.
+Current health lives in the **state machine**, not in the durable event archives. This doc exists
+because the codebase makes the wrong choice easy to reach for.
+
+## The rule
+
+> A health light reads the **live open states** of the relevant state machine.
+> It never reconstructs health from — or writes back into — a durable archive
+> (`failoverHistory`, `Crashes`, logs).
+>
+> One light = one place, one helper, one live signal.
+
+## The pipeline
+
+```
+SetState("WARN0186", state.State{ErrType, ErrDesc, ErrFrom:"REJOIN", ServerUrl})  // raised each tick
+        │   (a state not re-raised next tick self-clears; PreserveState keeps one across ticks)
+        ▼
+Cluster.StateMachine  ──►  GetOpenErrors()/GetOpenWarnings()/GetOpenInfos()
+        │
+        ▼
+GET /api/clusters/{name}/topology/alerts  ──►  { errors, warnings, infos } of { number, desc, from }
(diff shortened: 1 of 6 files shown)


svaroqui:
@claude please re-review — this PR was reworked after the first review. It is now **frontend-only**: the crash pill blinks on the cluster's **live HA health** (open `from:"REJOIN"` states in `clusterAlerts` — WARN0184/0185/0186/0187, self-clearing each tick), instead of reconstructing state from the crash archive. The earlier backend changes (`isRejoinResolved` + reconciling/rewriting `failoverHistory` records to `recovered`) were reverted, so both prior findings — the `stateSlaveErr` false-positive and the dropped `deltaAnalyzed && !deltaFlashable` alert — no longer apply. Please verify: (1) `crashRejoinNeedsAttention` correctly keys on live REJOIN states and nothing else; (2) `clusterAlerts` is populated wherever the Navbar pill renders; (3) the HADetail dead-code removal is truly unused; (4) the backend is unchanged vs develop.

PR #1622, opened 2026-07-22
fix(backup+config): abort backup on dead-volume stall; persist per-cluster settings across restart

svaroqui:
## Incident

A logical (`mariadb-dump` → `splitdump`) backup was running to a **mounted backup volume**; the mount was lost mid-backup. The database was never affected, but repman went sideways: the backup hung for 1h+, the source's `--single-transaction` read view stayed open (pinning InnoDB purge), and the GUI lit a **STALLED** pill that kept blinking **even after both OS processes were killed by hand**.

Diagnosed from a `/debug/pprof/goroutine` capture (two 30s-apart dumps): the monitoring loop was *advancing*, not deadlocked — three defects stacked.

## Root cause

1. **Dead output volume hangs the backup forever.** The hung mount blocks `splitdump`'s writes → back-pressures `mariadb-dump` and repman's read loop → `JobBackupMysqldump` parks on `wg.Wait()` with **no timeout**.
2. **The stuck backup drags monitoring** — job-state checks run inline on the `Ping`/`TopologyDiscover` path (documented; structural fix is follow-up).
3. **An unconditional `time.Sleep(3s)`** in `JobsCheckFinished` taxed every monitoring tick for every server.

Killing the processes didn't clear the pill because the hung goroutine never returned, so the deferred `InLogicalBackup` clear never ran — a pure in-memory ghost (the DB `jobs` table was already clean).

## Fix

- **Write-stall watchdog** (defect #1): track bytes flowing through the read loop; if none flow for `backup-write-stall-timeout` (default **300s**, `0` disables), cancel the already-cancellable `dumpCtx` — which kills **both** the dump and splitdump subprocesses (the cancel infra already existed at `srv_job_backup.go:2061`). The backup returns a clear stall error instead of hanging; unwinding runs the defers → `InLogicalBackup` clears, source transaction releases. **One fix resolves the hang, the pinned source transaction, and the ghost STALLED state.** It's a *stall* timeout, not a total-time cap — long healthy dumps are unaffected.
- **Sleep guard** (defect #3): the `time.Sleep(3s)` in `JobsCheckFinished` now runs only when there was a finished job to flush.

## Tests

- `srv_job_backup_stall_test.go` unit-tests the standalone watchdog: cancels on stall, never cancels while progressing, no-ops when disabled (`0`). `go build`/`go vet`/`go test ./cluster ./config ./server` green.
- **Controllable-mount reproduction** (loopback fs + lazy-unmount / `losetup`-detach) documented in `doc/implementation/cluster/BACKUP_DEAD_VOLUME_STALL.md` — the faithful way to validate the hung-mount variant before trusting it in production.

## Follow-ups (documented, not in this PR)

- Move job-state checks off the synchronous `Ping` path (defect #2) so no backup work can stall topology discovery.
- Expose `backup-write-stall-timeout` in the GUI Backup settings panel.

### DIFF
--- a/cluster/cluster.go
+++ b/cluster/cluster.go
@@ -348,6 +348,7 @@ type Cluster struct {
 	SstAvailablePorts                   map[string]string           `json:"sstAvailablePorts" groups:"web"`
 	InPhysicalBackup                    bool                        `json:"inPhysicalBackup" groups:"web"`
 	InLogicalBackup                     bool                        `json:"inLogicalBackup" groups:"web"`
+	LastConfigSaveToDisk                time.Time                   `json:"lastConfigSaveToDisk" groups:"web"` // when the datadir <cluster>.toml was last actually written (config persistence observability)
 	InBinlogBackup                      bool                        `json:"inBinlogBackup" groups:"web"`
 	InResticLogicalBackup               bool                        `json:"inResticLogicalBackup" groups:"web"`
 	InResticPhysicalBackup              bool                        `json:"inResticPhysicalBackup" groups:"web"`
@@ -1895,6 +1896,13 @@ func (cluster *Cluster) SaveConfigFile() (bool, error) {
 	s.WriteTo(&buf)
 
 	has_changed, err = cluster.saveConfigArtifact(filePath, buf.Bytes(), "saved-"+cluster.Name, true)
+	if err == nil && has_changed {
+		// Stamp the last time the persisted cluster config actually hit disk.
+		// Exposed via the API and GUI so operators (and the config-persistence
+		// regtest) can tell "changed in memory" from "written to the datadir
+		// <cluster>.toml that startup reads back".
+		cluster.LastConfigSaveToDisk = time.Now()
+	}
 	return has_changed, err
 }
 
--- a/cluster/srv_job.go
+++ b/cluster/srv_job.go
@@ -1127,8 +1127,15 @@ func (server *ServerMonitor) JobsCheckFinished(conn *sqlx.Conn) error {
 		server.SetNeedRefreshJobs(true)
 	}
 
-	//Wait for debug sent via API
-	time.Sleep(3 * time.Second)
+	// Wait for the writelog API to flush a finished job's status before we log it
+	// — but ONLY when there was actually a finished task. This runs on the
+	// monitoring hot path (Ping → Refresh → JobsCheckStates), so an unconditional
+	// sleep taxed every tick for every server even with no finished job, which
+	// (compounded by a stuck backup) is enough to trip the global heartbeat stall
+	// detector. Sleep only when there is something to flush.
+	if len(logs) > 0 {
+		time.Sleep(3 * time.Second)
+	}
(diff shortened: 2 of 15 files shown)


svaroqui:
## Added: config-persistence fix (same branch, one client test cycle)

Second, distinct bug found while diagnosing the same client incident: **custom per-cluster settings (e.g. `backup-mysqldump-options` with `--skip-ssl`) revert after a repman restart**, even though they're saved to the datadir `<cluster>/<cluster>.toml` `[saved-<cluster>]` and git (confirmed on the repo's master branch).

**Root cause — save/load gate asymmetry:** the *save* path is gated on `conf.ConfRewrite` (the `monitoring-save-config` pflag, default **true**), so saves happen. The *startup load* gate read `firstRead.GetBool("default.monitoring-save-config")`, which returns **false when the key is absent** from the config file — so `InitConfig` skipped loading the datadir saved overlays entirely. `true` for saving, `false` for loading. (Runtime reload via `ReconstructLiveClusterConfig` isn't gated this way, which is why it preserved the value while a full restart lost it.) OSC default for `monitoring-save-config` is `true`.

**Fix:** `monitoringSaveConfigEnabled()` defaults to the flag value when the key is absent, only using the file value when explicitly set — load gate now matches save gate. Explicit `= false` still disables.

**Observability:** `Cluster.LastConfigSaveToDisk` (API `lastConfigSaveToDisk`) stamped when the datadir `<cluster>.toml` is actually written — distinguishes 'changed in memory' from 'written to the file startup reads'.

**Tests:** `server_reload_test.go` (gate default + saved-option survives reload), `regtest testConfigPersistBackupOption` (live save round-trip). Doc: `doc/implementation/config/CONFIG_PERSISTENCE_SAVE_LOAD_GATE.md`.

Bundled here deliberately so one client restart validates both this and the backup-stall fix.

svaroqui:
Thanks for the review — addressed the follow-ups, with #1 done as a real fix (the must-do), not just a doc note:

1. **D-state / hard-hung mount (the must-do):** `SIGKILL` can't free a subprocess wedged in uninterruptible sleep, so the post-cancel `wg.Wait()`/`splitDumpPipeline.wg.Wait()` could still hang forever after cancellation. Added `boundedWait`: once the watchdog fires, we wait at most `backupStallLeakGrace` (60s) for the reader/pipeline goroutines to unwind; if they don't, the backup **returns the stall error and leaks** the unkillable goroutine instead of hanging. So the backup call always unwinds now (clears `InLogicalBackup`, releases the source transaction) even in the worst case — with a doc callout that this is best-effort for the OS process, not a hard guarantee.
2. **Config sample:** added `backup-write-stall-timeout` to `etc/config.toml`.
3. **Negative value:** now logs a warning (distinguishes intentional-disable from a typo).
4. **Detection latency:** documented (~`stallTimeout + checkInterval`, ~375s at the 300s default).

New test `TestBoundedWait_LeaksStuckGoroutineAfterStall` covers the leak-after-stall path. Build/vet/tests green.

PR #1625, opened 2026-07-22
harden(auth): bind API auth method to password presence

svaroqui:
Makes the API authentication model explicit and consistent across the local-password and OIDC paths, which share one `APIUsers` map.

## Model
- **Password present → local account:** authenticated by password only; SSO refused.
- **No password → SSO identity:** authenticated by OIDC only; local password refused.
- **Owner (`Cloud18GitUser`):** the one SSO account that legitimately carries a credential — exempt.

## Changes
- `cluster.IsValidACL`: OIDC refuses a password-protected account (except the owner); the password path refuses an empty stored or submitted password. `GetAPIUser` (gRPC) gets the same empty-password guard.
- OIDC callback (`server/api.go`): an SSO identity matching a password-protected local account is denied and logged (`api_sso_local_collision`) — loud, so an attempt to authenticate against a reserved local name (`admin`/`system`/`dba`) is visible in the security log rather than silently ignored.

## Why
Binding the auth method to password presence keeps a same-named SSO identity from riding a local account's ACL, and stops a passwordless entry (an SSO-provisioned peer, or a hand-written passwordless `api-credentials` line) from being accepted on the local password door.

## Migration
- A local API account **must** have a password (a passwordless entry is now SSO-only).
- Grant SSO peers by **email only** (no `:password`).
- Default `admin:repman` and normal password-protected accounts are unaffected.

## Tests
`cluster/cluster_acl_auth_guard_test.go` — correct/blank/wrong password locally, blank-password rejected for a passwordless account, SSO allowed for passwordless, SSO refused for a password-protected local account, owner exemption. Build/vet/existing ACL+auth suites green.

Model doc: `doc/implementation/security/API_CREDENTIAL_AUTH_MODEL.md`.

### DIFF
--- a/cluster/cluster_acl.go
+++ b/cluster/cluster_acl.go
@@ -7,6 +7,7 @@
 package cluster
 
 import (
+	"crypto/subtle"
 	"fmt"
 	"slices"
 	"sort"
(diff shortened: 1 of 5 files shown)


svaroqui:
Thanks for the review — addressed the follow-ups in the latest commit:

1. **Duplicated invariant** → extracted `cluster.IsLocalOnlyAccount(user)` as the single source of truth; both `IsValidACL`'s OIDC branch and the `server/api.go` collision guard now call it, so they can't drift.
2. **Event naming overlap** → added a cross-reference comment (and a doc note) distinguishing `api_sso_local_collision` from the pre-existing `api_login_local_sso_mismatch`.
3. **Security-log catalog** → added `api_sso_local_collision` to `SECURITY_LOGGING.md` (with the cross-reference to `api_login_local_sso_mismatch`).
4. **Constant-time compare** → `IsValidACL` and `GetAPIUser` now use `crypto/subtle.ConstantTimeCompare` instead of `==`.

Left as-is (agree they're out of scope for this PR):
5. **`GetAPIUser` raw vs `IsValidACL` decrypt** — confirmed intentional: the gRPC caller (`repmanv3.go`) passes the already-resolved password, whereas the HTTP login path submits the raw credential that `GetDecryptedPassword` normalizes. Not a latent bug, but I'll keep an eye on it.

Also added `TestIsLocalOnlyAccount`. Build/vet/auth tests green.

PR #1626, opened 2026-07-22
fix(monitor): run logical reseed off the tick (unblock monitor during restore)

svaroqui:
## Problem (confirmed from a live goroutine profile during a restore)

A splitdump/mysqldump **reseed runs on the cluster monitor tick goroutine**: `Run → tickBody → StateProcessing (cluster.go:1258) → ProcessReseedLogical → splitdump.Restore`. So for the entire multi-minute restore the cluster monitor loop is **frozen** — no health checks, **no failover even if the master dies mid-reseed** — and two concurrent slave reseeds **serialize** on the one tick goroutine (so "restore 2 slaves at once" isn't actually concurrent, and freezes monitoring the whole time).

## Fix

Dispatch the reseed via `cluster.trackTickGoroutine(...)` — **off the tick** — mirroring the WARN0111 async reseed path immediately below it. Dedup is already safe:
- `ProcessReseedLogical` claims the per-server `IsReseeding` flag atomically (`TrySetInReseedBackup`), so a later tick that still sees WARN0075 can't start a second reseed for the same server.
- **rejoin** (`srv_rejoin.go:482`) and **switchover staging** (`cluster_staging.go:449`) already refuse a reseeding server.
- Failover is master-triggered, so a slave reseed can't trip it.

## Why this is a net safety gain

Today the frozen monitor can't fail over even if the **master** dies during a slave reseed. After the fix the monitor keeps running, so that failover works — and the two slaves genuinely reseed in **parallel**. The only new exposure is cosmetic WARN states on the reseeding slave (its replication is stopped mid-restore); the structural re-actions are already `IsReseeding`-guarded.

## Testing

Builds/vets clean; cluster reseed/tick/state tests pass. **This is an orchestrator hot-path change — please run it on the OpenSVC topology matrix before merge** (concurrent 2-slave reseed + master-failover-during-reseed are the key scenarios).

### DIFF
--- a/cluster/cluster.go
+++ b/cluster/cluster.go
@@ -1239,6 +1239,12 @@ func (cluster *Cluster) StateProcessing() {
 					}
 					cluster.LogModulePrintf(cluster.Conf.Verbose, config.ConstLogModGeneral, config.LvlErr, "Fail of processing reseed for %s: %s", servertoreseed.URL, err)
 				}
+				// Reconcile a rejoin-armed reseed with the real outcome (deferred at
+				// arm time). See the WARN0075 handler below for the ordering rationale.
+				if servertoreseed.reseedFromRejoin.Load() {
+					cluster.finishRejoin(servertoreseed.URL, rejoinResultOf(err))
+					servertoreseed.reseedFromRejoin.Store(false)
+				}
 			}
 
 			if s.ErrKey == "WARN0075" && servertoreseed != nil {
@@ -1267,12 +1273,21 @@ func (cluster *Cluster) StateProcessing() {
 				srvReseed := servertoreseed
 				reseedTask := task
 				cluster.trackTickGoroutine(func() {
-					if err := srvReseed.ProcessReseedLogical(reseedTask); err != nil {
+					err := srvReseed.ProcessReseedLogical(reseedTask)
+					if err != nil {
 						srvReseed.JobsUpdateState(reseedTask, err.Error(), 5, 1)
 						if srvReseed.HasReseedingState(reseedTask) {
 							srvReseed.SetInReseedBackup("")
 						}
 						cluster.LogModulePrintf(cluster.Conf.Verbose, config.ConstLogModGeneral, config.LvlErr, "Fail of processing logical reseed for %s: %s", srvReseed.URL, err)
+					}
+					// If this reseed was armed by a rejoin, reconcile the rejoin record
+					// with the REAL restore outcome (finishRejoin was deferred at arm
+					// time). finishRejoin first (→ rejoinAlreadyAttempted true), then
+					// clear the marker, so the one-shot never has an open gap.
+					if srvReseed.reseedFromRejoin.Load() {
+						cluster.finishRejoin(srvReseed.URL, rejoinResultOf(err))
+						srvReseed.reseedFromRejoin.Store(false)
 					}
 				})
 			}
--- a/cluster/cluster.go
+++ b/cluster/cluster.go
@@ -1255,14 +1255,26 @@ func (cluster *Cluster) StateProcessing() {
 					})
 				}
 
-				err := servertoreseed.ProcessReseedLogical(task)
-				if err != nil {
-					servertoreseed.JobsUpdateState(task, err.Error(), 5, 1)
-					if servertoreseed.HasReseedingState(task) {
-						servertoreseed.SetInReseedBackup("")
+				// Run the logical reseed OFF the monitor tick. A splitdump /
+				// mysqldump restore takes minutes; running it inline here froze
+				// the whole cluster monitor loop (no health checks, no failover —
+				// not even if the master died) for the entire restore, and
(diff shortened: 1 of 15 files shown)


svaroqui:
@claude please review this update.

New in this push (commit 0a42b6d29 — "rejoin: async everywhere + reconcile the async-reseed record"):

1. **All rejoin edges now async.** `cluster_topo.go` extra-master caller changed from `extra.RejoinMaster()` to `go extra.RejoinMaster()` — it was the last synchronous rejoin caller, and it ran on the monitor tick, which is why the reseed had to be decoupled off-tick.

2. **Async-reseed record reconciliation.** With the logical/physical backup reseed only arming WARN0075/WARN0074 and returning while the restore runs off-tick, `finishRejoin` was recording *success at arm time*, before the restore ran. Now:
   - `recordOrDeferRejoin()` records immediately for a synchronous restore (RejoinDirectDump), or sets `reseedFromRejoin` and defers when an async reseed is in flight;
   - the WARN0074/WARN0075 completion handlers reconcile `finishRejoin` with the real `ProcessReseed*` outcome (finishRejoin then clear, so the one-shot has no gap);
   - `RejoinMaster` holds the one-shot while a reseed is in flight (`HasAnyReseedingState() || reseedFromRejoin`).

Please focus on: the re-entrancy/one-shot correctness of the `reseedFromRejoin` marker vs `rejoinAlreadyAttempted`, any race between the reseed clearing `IsReseeding` and `finishRejoin` landing, and whether the physical (WARN0074, inline) vs logical (WARN0075, off-tick) handlers both reconcile safely. No retry semantics were changed — the record is one-shot either way, only made honest for the async paths.

Docs updated in `UNIFIED_REJOIN_DESIGN.md`: execution model, full rejoin workflow (triggers/gates/per-path states), and the WARN0189 reseed-progress coverage gap.

svaroqui:
@claude thanks — both criticals were real and are fixed in `0f4fc6b24`. Your root cause was exactly right: I hooked the reconciliation to the WARN-handler *return values*, but every reseed method arms a detached goroutine and returns `nil` first.

**#1 (physical records false success) & #2 (direct-mysqldump never clears the marker → RejoinMaster permanently disabled)** — fixed together by removing the WARN0074/WARN0075 handler reconciliation entirely and reconciling at the **one signal every method shares: `IsReseeding` clearing**, from **observed health**:

- `reconcileDeferredRejoinReseeds()` (`srv_lostevents.go`, per tick, before `assertRejoinResultStates`): for a server with `reseedFromRejoin` set and `IsReseeding` **cleared** → `finishRejoin(success)` if it's a healthy slave of the current master, else `failed`; then clear the marker.
- A node still establishing replication right after the reseed self-corrects: `assertRejoinResultStates` already rewrites `failed → recovered` once it's a healthy slave, so no stale failure sticks. That reuse is why I went observed-health instead of trying to thread a real completion callback through `WaitAndSendSST`/`SSTRunSender`/`JobRejoinMysqldumpFromSource`.
- This also covers the direct-mysqldump path uniformly (it has no WARN state), so #2 can't recur.

**#3 (unsynchronized `cluster.Crashes`)** — acknowledged for the `-race` matrix run. Note this change *reduces* concurrent writers for the reseed-completion path: that `finishRejoin` now runs on the locked monitor tick (inside `TopologyDiscover`) instead of in detached reseed goroutines. The remaining concurrent caller is `go RejoinMaster` for the synchronous exits; a dedicated crashes mutex is a good follow-up but out of scope here.

Also folded in a generic `WARN0189` "rejoin reseed in progress, started T" (from `rejoinReseedStart`) so a rejoin is always visible in the GUI even for methods without byte instrumentation — the outcome landing at completion is the visible proof the reconcile works. Please re-review.

svaroqui:
@claude re-review please — addressed the two stale comments from your last pass, and this push adds the GUI proof-of-work.

**Stale comments (5488c4eb):** the guard comment (srv_rejoin.go:82) and `recordOrDeferRejoin`'s doc now point at `reconcileDeferredRejoinReseeds` / observed-health, not the removed WARN0074/WARN0075 handler reconcile.

**Live reseed/rejoin progress GUI (7cc9d380):**
- Backend: `ServerMonitor.GetReseedProgress()` (`cluster/restore_progress.go`) snapshots the in-flight restore — bytes/total/percent for byte-instrumented methods, or a generic timer from `rejoinReseedStart` — attached to `GET /topology/servers` as `server.reseedProgress` (nil when idle) in `server/api_cluster.go`.
- Frontend: a Navbar 'Reseed' pill appears while any server restores → opens `ReseedProgressModal` (Chakra `Progress` bar when byte-instrumented, indeterminate + 'started T' line otherwise). `crashPill` now excludes WARN0189 so a normal reseed no longer red-blinks the crash pill. Pure logic in `utility/reseedProgress.js` with node tests (18 pass); `npm run build` green.

Please review the GUI slice (esp. the `reseedProgress` API shape and the WARN0189 crash-pill exclusion) plus confirm the comment fixes.

PR #1627, opened 2026-07-23
splitdump: native Go transactional restore (replace mysql --force) + reseed progress

svaroqui:
## Why
A splitdump reseed silently lost rows: shards were piped to `mysql --force`, which skips a failing statement and still exits 0. `executeMysqlRestoreContext` only read stderr on a non-zero exit, so any shard that errored (typically the first shard — it carries `LOCK TABLES`/`DISABLE KEYS` and loses the deadlock race under N-way parallel load) was reported as a **successful** restore with its rows dropped. The split and the manifest were proven correct; the loss was entirely at the `mysql --force` execution layer.

## What
1. **Native Go transactional loader** (`9b477fc5b`): each shard runs over a **dedicated pinned `*sqlx.Conn` per worker** (never the shared pool). INSERT/REPLACE are batched into transactions and **retried on deadlock (1213) / lock-wait-timeout (1205)**; any other error is returned so the restore **fails loud**. LOCK/UNLOCK TABLES and ALTER..DISABLE/ENABLE KEYS are dropped; DELIMITER is honored. Binlog on/off follows the master/slave decision via `GetConnNoBinlog`; reset uses the version-aware `server.ResetMaster()`.
2. **Manifest-completeness test** (`3e51c5d2b`).
3. **Reseed progress (WARN0189) for splitdump** (`eb1bd9542`): the multi-file restore now accumulates streamed bytes into one counter, so a splitdump reload shows the same live `<streamed> out of <total>` line per tick as the monolithic path (was a silent 'processing').
4. **Remove the dead `mysql --force` CLI path** (`e2a6335a6`) + its tests.

## Review focus
- transaction/retry correctness in `execSplitdumpBatch` (deadlock replay, non-retryable → abort);
- statement segmentation in `streamSplitdumpStatements` (DELIMITER, LOCK TABLES stripping, INSERT batching);
- per-worker connection lifecycle (dedicated conns, no pool multiplexing);
- progress byte accounting across concurrent shard readers.

@claude please review.


svaroqui:
@claude thanks — fixed the `continueOnError` regression and added the parser tests you asked for, in `73e52458e`.

**The regression (blocking):** you were exactly right — `continueOnError` reached `execSplitdumpSingle` but not `execSplitdumpBatch`, so `mysql.system-all` INSERTs went through the transactional batch path and would hard-abort the whole restore on the first conflicting seed row. Fixed: when `continueOnError`, INSERT/REPLACE now run **individually** (`single`, coe=true) so a bad row is skipped like the old `--force`, instead of rolling back the transaction.

**Refactor for testability** (your test ask): split the streamer into pure pieces —
- `classifySplitdumpStatement` (skip / insert / other),
- `forEachSplitdumpStatement` (DELIMITER-aware segmentation, no DB),
- `planAndExecSplitdump` (routing) driven by an injectable `splitdumpExecutor`.

New tests (no live DB, all green):
- `TestForEachSplitdumpStatement` — DELIMITER trigger body, comment skipping, terminator stripping, embedded `;` in a value, trailing no-terminator statement.
- `TestClassifySplitdumpStatement` — LOCK/UNLOCK/DISABLE-KEYS→skip, INSERT/REPLACE→insert (case-insensitive), else→other.
- `TestPlanAndExecSplitdumpContinueOnError` — **the regression guard**: continueOnError → 2 individual execs, 0 batches; normal → 1 batch of 2, 0 singles.
- `TestPlanAndExecSplitdumpDropsAndFlushes` — dropped statements never execute; a non-INSERT flushes the pending batch first.

**Minor notes:** added the parser-assumption comment (per-line `HasSuffix` is mysqldump-specific, not a general tokenizer) and the per-worker `*sqlx.DB` rationale. Left `DISABLE KEYS` on `Contains` (it arrives wrapped in `/*!40000 ALTER TABLE ... */`, so it isn't at statement start) — noted in the classifier comment. Please re-review.

svaroqui:
Addressed the comment-placement nit in `92b6d4dd4`: removed the orphaned function doc that was left above `type splitdumpStmtKind` and gave the `streamSplitdumpStatements` wrapper its own concise doc. Comment-only; build + `go test ./cluster/ -run 'Splitdump'` green. Thanks for the review — no other changes.

PR #1630, opened 2026-07-28
feat(logs): add configurable fetched DB log retention

caffeinated92:
## Summary
- add dedicated fetched DB log retention settings and UI/API wiring
- add optional backup-backed storage for fetched DB logs with safe legacy log migration
- unify scheduler, API receive-task, slow-log export, and tailers on the same DB log path helper
- reuse the shared s18log/lumberjack writer for fetched DB log rotation and add focused tests

## Why this PR exists
Fetched DB logs currently have inconsistent behavior and location handling:
- scheduled fetch jobs and tailers historically used the legacy per-server cluster path
- API-mode receive-task could target a different location
- runtime DB log rotation relied on ad hoc behavior instead of the shared logging backend
- operators often keep larger partitions behind backup-backed storage, but there was no safe way to move fetched DB logs there while preserving old files

This PR creates one configurable model for fetched DB logs:
- `db-log-rotate` enables repman-managed rotation for fetched DB logs only
- `db-log-on-backup-storage` lets operators opt into backup-backed storage while keeping legacy behavior by default
- old legacy DB logs are migrated safely without overwriting destination files
- tailers are restarted when the storage location flips so they do not keep following stale paths

## Validation
- `go test ./cluster -run 'Test(MigrateDBLogsToBackupStorage|EnsureDBLogsMigrated|RenameOrCopyFile|RestartLogTailers)' -count=1`
- `go test ./utils/s18log -run 'TestNewRotateWriter' -count=1`


caffeinated92:
@claude review the changes. And check if we still have other bugs in this PR

PR #1645, opened 2026-08-03
fix(arbitrator): fail over backend DB connections

caffeinated92:
## Why
When the arbitrator database backend goes down, the arbitrator should not go down with it. In deployments with multiple backend database hosts, we need the arbitrator to keep serving and fail over to a secondary backend if the current or primary host is unavailable.

## What changed
- stop treating arbitrator DB connection loss as a fatal startup condition
- add reconnect logic for the arbitrator DB handle
- try configured MySQL backend hosts in order, preferring the last known good host first
- add a short per-host connect timeout so a dead primary does not stall failover
- keep heartbeat table creation best-effort so least-privilege accounts can still reconnect

## Validation
- go test -tags arbitrator ./arbitrator ./config

### DIFF
--- a/arbitrator/arbitrator.go
+++ b/arbitrator/arbitrator.go
@@ -201,6 +201,26 @@ var (
 	reconnectCooldown    = 2 * time.Second
 )
 
+// defaultArbitratorConnectTimeoutSeconds is both the flag default and the
+// fallback used whenever arbitrator-connect-timeout is non-positive.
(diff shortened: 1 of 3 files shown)


caffeinated92:
@claude review staged changes, make sure we don't introduce bugs

caffeinated92:
@claude review and make sure we don't introduce bugs

PR #1652, opened 2026-08-05
Fix critical MariaDB PFS digest capture that hammers client DBs (#1651)

### LINKED ISSUE #1651: PFS Top-Queries digest capture is pathologically expensive and can't be disabled at runtime
## Summary

When the **Top Queries** view is opened in the GUI, repman starts running its Performance-Schema digest-capture query (`GetQueries`) against the monitored server on **every monitoring tick**. On a busy production server this statement is very expensive (CPU + network) and can starve application queries — and there is **no way to turn it off at runtime**: none of the GUI monitoring toggles stop it; only a config reload/restart does.

## Observed

- **Only reproduces after opening the Top Queries view in the GUI.** Once it starts it keeps running every tick and cannot be stopped from the GUI.
- The offending statement is repman's own (tagged `/*replication-manager*/`):

```sql
SELECT /*replication-manager*/
    A.digest, '' as query,
    COALESCE((
        SELECT B.SQL_TEXT
        FROM performance_schema.events_statements_history_long B
        WHERE B.DIGEST = A.DIGEST
          AND B.SQL_TEXT IS NOT NULL
          AND B.SQL_TEXT NOT LIKE '%replication-manager%'
        LIMIT 1
    ), '') as sample_query,
    A.digest_text, A.LAST_SEEN, COALESCE(A.SCHEMA_NAME,''),
    ... (full digest stats) ...
FROM performance_schema.events_statements_summary_by_digest A
WHERE A.digest_text is not null;
```

- One execution runs one-at-a-time (an in-progress guard holds), but on a loaded server a single execution runs long and competes with application traffic.

## Root cause (verified in source)

1. `cluster/srv.go:1207` launches `go server.GetPFSQueries()` **unconditionally every monitoring tick** — no gate at the launch site.
2. Inside, it is gated **only** by `monitoring-performance-schema` (`MonitorPFS`, `cluster/srv_get.go:524`). It is **not** gated by `monitoring-queries` **nor** `monitoring-processlist-information-schema`, so toggling those has no effect on this query.
3. There is **no runtime toggle for the master `monitoring-performance-schema`**: the switch handler (`server/api_cluster.go:2703-2712`) only has cases for the *sub*-flags (`-mutex`, `-latch`, `-memory`, `-instruments`, `-queries`). Changing `monitoring-performance-schema` in the GUI persists to file but never updates the **running** value — it only takes effect after a config reload/restart. So the query cannot be disabled live.
4. `GetQueries` (`utils/dbhelper/performance.go:208`) on MariaDB issues a **per-digest correlated subquery** into `events_statements_history_long` for **every** row of `events_statements_summary_by_digest`, with **no `LIMIT`/threshold**, every tick — O(digests × history). It also returns full `digest_text` + `sample_query` for all rows, shipping a large result set over the network each cycle.

## Impact

On a busy server this burns DB CPU and network bandwidth and can stall application queries. Because there is no runtime off-switch, an operator cannot mitigate it without restarting repman.

## Proposed fix

1. **Gate the launch** at `srv.go:1207` on the live `MonitorPFS` (and the in-progress guard) so nothing spawns when it is disabled.
2. **Add a runtime toggle** (`SwitchMonitorPFS` + a `monitoring-performance-schema` case in the switch handler) so it can be turned off without a restart.
3. **Make the query cheap:** drop the per-digest correlated subquery (use the existing lazy `GetSampleQueryFromPFS` on demand / on drill-in), add a `SUM_TIMER_WAIT` threshold + `LIMIT`, throttle the fetch cadence, and add a context/timeout so a single slow execution can never hang holding the guard.

svaroqui:
Fixes #1651.

The GUI Top-Queries PFS digest capture (`dbhelper.GetQueries`) ran a **per-digest correlated subquery** into `events_statements_history_long` — `O(digests × history)`, and MariaBs PFS is unindexable — on **every monitoring tick**, and had **no live off-switch**. Confirmed live on a partner: the master carried **4014 digests** (legacy app, `IN(...)` / non-parameterized SQL → digest explosion) → ~**40M row-scans every ~2s**, starving application queries.

## Fix (repman-side only; no DB config changes)
- **`GetQueries` (MariaDB):** replace the correlated subquery with a **non-correlated derived table** (`GROUP BY DIGEST`) joined once. MariaDB materializes it once and **auto-keys** it (`derived_with_keys`) → `O(digests + history)`. A projection-list scalar subquery can’t be decorrelated by the optimizer, so we hand it a join it can index/hash. **Proven live: `EXPLAIN` shows `<derived2> key=key0`, 0.356s vs the burn.** MySQL path unchanged (reads `QUERY_SAMPLE_TEXT`, no join, zero MariaDB-specific SQL).
- **Dedicated connection + session-scoped speed knobs** (`optimizer_switch=derived_with_keys`, `join_cache_level` — MariaDB-only, flavour-gated) + a guard timeout. We force what makes *our* query fast at the session level and **never depend on or mutate the client’s global config**.
- **Throttle:** launch the capture every **N monitoring ticks** (modulo idiom, same as `ResticFetchRepo` / `BackupReconcileInterval`), config `monitoring-performance-schema-queries-interval` (default 30). No more every-tick.
- **Live off-switch:** gate the launch on live `MonitorPFS` + add `SwitchMonitorPFS` (+ the `monitoring-performance-schema` API case) so the GUI toggle actually disables it live (it had no live switch before — turning it off never reached the running loop).

## Verification
- `EXPLAIN` + timing run on the real 4014-digest master (auto-key confirmed, `derived_with_keys=on` / `join_cache_hashed=on`).
- `./utils/dbhelper/` builds; gofmt-clean. (Full-repo build to be confirmed in CI — this checkout has unrelated pre-existing module breakage.)

**North star: a stale advisory is acceptable; hammering the client’s DB is not.**


svaroqui:
@claude please do a fresh full re-review of the current state of this PR. Since the initial review I have pushed: (1) rewritten the stale GetQueries doc comment; (2) made the dedicated-connection isolation real (restore optimizer_switch/join_cache_level/group_concat_max_len to DEFAULT before the pooled connection is released); (3) extracted a pure pfsQueriesSQL and added unit tests (performance_pfs_test.go); (4) kept monitoring-performance-schema-queries-interval (0=disabled, tick modulo, consistent with backup-reconcile-interval) with clarified help text; (5) added the missing master monitoring-performance-schema toggle + interval field to MonitoringSettings.jsx. Please confirm whether it is now clean to merge, or flag anything remaining.

svaroqui:
Fresh re-review verdict: **clean to merge**. The reviewers only caveat was that its sandbox could not run go/npm — verified locally this session: `go build ./cluster ./server ./config ./utils/dbhelper` all pass, `npm run build` (dashboard) passes, and the 3 `pfsQueriesSQL` unit tests pass. Merging.

PR #1654, opened 2026-08-06
Fix WARN0190/WARN0191 alert flap (rejoin-catalog states missing from pstates30)

svaroqui:
## Problem

`HasCatalogBackupForRejoin()` runs only every 30 ticks (`heartbeats%30==0`) and asserts **WARN0190/WARN0191**, but neither was listed in `pstates30`.

On the 29 intermediate ticks between checks, the state machine sees these states in `OldState` but not `CurState` and **resolves** them — then tick 30 re-raises them. The result is a continuous open→resolve→open **flap** that:

- spams the monitoring log (one line per transition), and
- fires **~10K Mattermost alerts/day/cluster** (every transition is an alert), making real alerts un-surveillable.

This is precisely the contract documented in `doc/monitoring.md`:

> For checks that run periodically (every N ticks), `PreserveState()` keeps their alerts visible on intermediate ticks.

A state raised by a `%N`-cadence check **must** be in `pstatesN`, or it flaps. WARN0190/WARN0191 were simply never registered.

## Fix

- **`cluster/cluster.go`** — add `WARN0190`/`WARN0191` to `pstates30` (with a comment tying them to the `%30` check).
- **`doc/implementation/cluster/MONITOR_LOOP_FLOW.md`** — the `%30` enumeration was also stale; added `HasValidBackup()` and `HasCatalogBackupForRejoin()` so the documented list matches the loop.
- **`cluster/cluster_bck_test.go`** — two regression guards:
  - `TestPstates30_IncludesRejoinCatalogStates` — membership guard.
  - `TestRejoinCatalog_WARN0191_PreservedAcrossIntermediateTicks` — reproduces the flap and proves `pstates30` preservation fixes it, with a control state (`WARN9999`, absent from `pstates30`) that must resolve — keeping the test sensitive so it fails if WARN0191 is ever removed.

## Scope / safety

Alerting/state-preservation only. No change to rejoin, failover, or any orchestration behavior — the underlying `HasCatalogBackupForRejoin()` logic is untouched; this only stops its state from self-resolving between checks.

## Test

```
go build ./cluster/            # OK
go test ./cluster/ -run 'TestPstates30_IncludesRejoinCatalogStates|TestRejoinCatalog_WARN0191_PreservedAcrossIntermediateTicks'   # PASS
```

### DIFF
--- a/cluster/cluster.go
+++ b/cluster/cluster.go
@@ -926,6 +926,7 @@ var pstates30 = []string{
 	"WARN0141", "WARN0142", "WARN0143", "WARN0150", "WARN0151", // Tresholds
 	"WARN0153", // Job related
 	"WARN0158", // Job secrets mismatch
+	"WARN0190", "WARN0191", // Rejoin catalog (HasCatalogBackupForRejoin runs %30)
 	"CREDIT01", // Credit related
 }
 
--- a/cluster/cluster_bck_test.go
+++ b/cluster/cluster_bck_test.go
@@ -2381,3 +2381,68 @@ func TestResticInitRepoWithOptions_ProceedsWhenBucketPreCheckFails(t *testing.T)
 		t.Fatal("expected init to still attempt and fail (no real restic binary here), not return nil early")
 	}
 }
+
+// TestPstates30_IncludesRejoinCatalogStates guards the contract documented in
+// doc/monitoring.md: a state whose check runs only every N ticks MUST be listed
+// in pstatesN, or it flaps open/resolve on every intermediate tick and storms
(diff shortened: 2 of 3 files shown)


svaroqui:
Addressed the doc nit in ddc30f1: removed the duplicate `HasValidBackup()` and repositioned `HasCatalogBackupForRejoin()` to match the real `%30` call order in `cluster.go` (`HasValidBackup → HasCatalogBackupForRejoin → CheckCanSaveDynamicConfig`), rather than after `CheckAllBackupEstimatedSize`. `HasValidBackup()` now appears exactly once. Fix + tests unchanged. Thanks for catching it.

Issue #1665, opened 2026-08-06
Self-heal oversized repman-side collected DB logs (reclaim a runaway 20G log automatically, safely)


svaroqui:
## Summary

Even with the inflation root causes fixed (#1661 dbjob re-stream, #1663 startup re-parse), a client can already be sitting on a multi-GB runaway collected DB log (crm/db1: ~12GB real / 179GB sparse `log_slow_query.log`). repman should **self-heal** that broken-deployment state on its own — detect the oversized collected log and reclaim the disk — instead of waiting for a human to intervene.

## Why this is safe NOW (and was not before)

This does **not** violate "never resolve an issue by deleting what triggers it":
- Before today, truncating that file would have masked a live bug and destroyed the only evidence of it.
- Now the cause is fixed (#1661 stops re-streaming the whole file, #1663 stops re-parsing it), so the accumulated bulk is confirmed *garbage from a fixed bug*, not live evidence.
- The self-heal still **preserves a bounded compressed sample** and **emits a visible state**, so nothing is hidden — it's remediation with an audit trail, not a silent truncate.

## Proposed behavior

- **When:** at startup (before the tailer attaches, in `NewLogTailer`) and optionally on a periodic check.
- **Detect:** a repman-side collected DB log file (`DBLogDir()/log_*.log`) exceeds a hard ceiling (independent of the `db-log-rotate` flag — this is repman's own housekeeping). Proposed ceiling: a plain absolute cap (e.g. 1 GB) via a named constant, easy to tune.
- **Act:** move the oversized file aside to a timestamped backup, immediately create a fresh empty file so the tailer attaches to it, then (in the background) gzip the backup and remove the uncompressed copy — reclaiming the disk while keeping a `.gz` for forensics.
- **Guard:** skip if the file is currently open for an SST receive (`IsFileOpenForSSTReceive`) to avoid racing an in-flight writer.
- **Make it visible:** emit a state/log at WARN so operators see "repman reclaimed an NN GB runaway collected log" (ties into the INFO/state work in #1662).

## Scope

- Independent of the deferred `db-log-rotate` flag semantics (that flag is about DB-**node**-side behavior; this is repman-side housekeeping).
- Complements: #1661 (stop the flood), #1663 (stop the re-parse), #1662 (surface state).

**Affected:** `cluster/srv.go` (`NewLogTailer`), a new self-heal helper, `config/error.go` (state code).

svaroqui:
[...]

Issue #1663, opened 2026-08-06
CRITICAL regression: repman-side stored DB logs bypass the always-on rotate+gz+retention → unbounded 20G log_slow_query.log


svaroqui:
## Summary

repman-side stored logs have **always** been rotated by size, retained by file count, and gzip-compressed (`.gz`). This is a long-standing, always-on guarantee. It is now **broken** for the logs repman fetches and stores from managed servers: the collected `log_slow_query.log` (and the other fetched DB logs) grow **unbounded and uncompressed**.

Observed on dbaas-fr-2: collected `log_slow_query.log` ≈ **12 GB real / ~179 GB apparent (sparse)** in repman's `Datadir/log`, never rotated, never compressed.

## The always-on policy (what should apply)

`server/server.go` wires every repman log through lumberjack with:
- `LogRotateMaxSize` (default 5 MB) — rotate by size
- `LogRotateMaxBackup` (default 7) — keep N files
- `LogRotateMaxAge` (default 7 d)
- `Compress: true` → `.gz`

This covers `replication-manager.log`, `carbonapi.log`, component logs, etc. — always, no flag.

## Root cause of the regression

The **fetched DB logs** (`log_error.log`, `log_slow_query.log`, `log_sql_error.log`, `log_audit.log`) — which are **repman's own stored copies**, living under `DBLogDir()` = `Datadir/log` — were split onto a *separate* writer path (`getDBLogRotatingWriter`) gated by `db-log-rotate`, which **defaults off**:

`cluster/cluster_sst.go:206`
```go
if !cluster.Conf.DBLogRotate {
    return cluster.SSTRunReceiverToFile(server, filename, ConstJobAppendFile, task) // append forever, no rotation, no compression
}
```

So this class of repman-side files silently dropped out of the always-on rotate+gz+retention guarantee → unbounded 20 G file.

## Semantic problem with the flag (must be fixed as part of this)

`db-log-rotate` reads as **"what repman does to the client database's own logs"** (DB-side) — and there the correct default is *don't touch them* (off). But the code uses this same flag to gate rotation of **repman's own stored copies**, which should **always** be rotated regardless. The flag is applying a "leave it alone" default to the **wrong target**.

## Fix direction

1. **repman-side stored logs are always rotated** — route the fetched DB log files through the existing always-on `LogRotate*` policy (size + count + `Compress: true`) unconditionally. No opt-out; repman must never keep an unbounded/uncompressed copy of anything it stores.
[...]

PR #1666, opened 2026-08-06
Fix slow-query-log inflation + generalize repman log rotate/compress + self-heal runaway logs

### LINKED ISSUE #1661: CRITICAL: dbjobs dblogfile() never persists stream offset → slow-query log re-streamed in full every run (100× inflation + monitor CPU spin)
## Summary

`dblogfile()` in `share/scripts/dbjobs_new.sh` **reads a streaming offset state file it never writes**. As a result the DB-side slow-query (and any `dblogfile`-collected log) is **re-streamed in full on every job run**, and repman appends each full copy to its collected log. The collected `log_slow_query.log` inflates without bound (~100×), which in turn drives a monitor-side fingerprint CPU spin.

Observed on production dev cluster **crm/db1** (`crm..db1.container.jobs`, s18-fr-4):
- DB source `slow-query.log` ≈ **124 MB** (stale, last real write Jul 26).
- repman-side collected `log_slow_query.log` ≈ **12 GB real / 179 GB apparent (sparse)**, still growing.
- `ls /var/lib/mysql/.system/jobs/*.state` → **no state file exists at all.**

## Root cause

`dblogfile()` (`share/scripts/dbjobs_new.sh`, lines ~1280–1314):

```sh
dblogfile() {
    local DBLOG="$1"; local JOB="$2"
    local STATEFILE="$TMP_DIR/${JOB}.state"
    ...
    local LAST_LINE=""
    [ -f "$STATEFILE" ] && LAST_LINE=$(cat "$STATEFILE")   # READS state
    local NEXT_LINE=1
    if [ -n "$LAST_LINE" ]; then
        if grep -Fxn -- "$LAST_LINE" "$DBLOG" >$TMP_DIR/match_pos; then
            NEXT_LINE=$(( $(cut -d: -f1 $TMP_DIR/match_pos | tail -n1) + 1 ))
        fi
    fi
    local TMPLOG="$TMPDIR/${JOB}.newlines"                 # BUG2: $TMPDIR (unset) not $TMP_DIR
    tail -n +"$NEXT_LINE" "$DBLOG" > "$TMPLOG"
    if [ -s "$TMPLOG" ]; then
        socat -u stdio TCP:$ADDRESS < "$TMPLOG" &>"$LOG_DIR/$JOB.process.out"
    else
        echo -n | socat -u stdio TCP:$ADDRESS &>"$LOG_DIR/$JOB.process.out"
    fi
}                                                          # <-- ends here; STATEFILE never written
```

- **Bug 1 (critical):** the function never writes `$STATEFILE` back after streaming. `grep` for any write to `STATEFILE` in the file = **0 hits**. So `$STATEFILE` never exists → `LAST_LINE=""` → `NEXT_LINE=1` → `tail -n +1` re-streams the **entire** file every invocation.
- **Bug 2 (latent):** line ~1303 uses `$TMPDIR` (unset) instead of the `readonly TMP_DIR`, so the newlines temp file lands at `/${JOB}.newlines` instead of under the jobs dir.

repman's receiver (`SSTRunReceiverToDBLogFile`, `ConstJobAppendFile` append mode) appends each re-streamed full copy → unbounded growth of the collected log. The monitor's `SlowLogWatcher` then fingerprints an ever-growing re-appended file → CPU spin.

## Impact

- **Every** managed DB, not just crm — `dbjobs_new.sh` is pushed to all DBs via the `jobs-upgrade` self-update, and crm is already running the latest (self-upgraded) copy, so the fix reaches every cluster.
- Runaway disk usage on the repman node (multi-GB/tens-of-GB collected logs).
- Wasted network + monitor CPU re-processing the same lines indefinitely.
- Slow-log data delivered to log plugins is massively duplicated → plugin metrics/counters are wrong.

## Proposed fix

After a successful stream, persist the last streamed line as the resume marker, and fix the temp-dir variable:

```sh
    local TMPLOG="$TMP_DIR/${JOB}.newlines"   # was $TMPDIR
    tail -n +"$NEXT_LINE" "$DBLOG" > "$TMPLOG"
    if [ -s "$TMPLOG" ]; then
        socat -u stdio TCP:$ADDRESS < "$TMPLOG" &>"$LOG_DIR/$JOB.process.out"
        tail -n 1 "$DBLOG" > "$STATEFILE"     # persist offset for next run
    else
        echo -n | socat -u stdio TCP:$ADDRESS &>"$LOG_DIR/$JOB.process.out"
    fi
```

(Consider a byte-offset marker instead of last-line matching if the source is rotated/truncated between runs, which would otherwise reset `grep -Fxn` to line 1.)

## Notes / scope

- Fix belongs on the orchestrator/dbjobs branch (pushed to all DBs) — not a direct `develop` commit.
- Cleanup of the already-inflated repman-side files and the missing `getDBLogRotatingWriter` size cap are **separate** follow-ups; this issue is only the dbjobs offset bug.
- No live data was truncated/deleted during investigation (read-only).

**Affected file:** `share/scripts/dbjobs_new.sh` — `dblogfile()`
**Verified on:** crm/db1 (`crm..db1.container.jobs`), active script self-upgraded 2026-08-06 14:26.

svaroqui:
## What & why

Fixes the runaway slow-query-log incident traced on crm/db1 (dbaas-fr-2): repman was inflating a collected DB log ~100× (124 MB source → 12 GB real / 179 GB sparse) and burning CPU re-processing it. Root-caused to **four independent problems**, each fixed here. All fixes cover **all four** collected DB log kinds (error, slow_query, audit, sql_error), not just slow query.

Closes #1661, #1663, #1665.

### 1. dbjob persists its stream offset (#1661)
`dblogfile()` in `share/scripts/dbjobs_new.sh` read `$STATEFILE` to resume streaming but **never wrote it back** → the state file never existed → `NEXT_LINE` stayed 1 → every run re-streamed the **entire** DB log from line 1, which repman appended (the actual flood). Now the last non-empty streamed line is persisted on a successful send. Also fixed `TMPLOG` using the unset `$TMPDIR` instead of `$TMP_DIR`.
> Propagates to every managed DB via the `jobs-upgrade` self-update.

### 2. Tailer follows from EOF, not the beginning (#1663)
`NewLogTailer` opened the collected logs with no `tail` `Location`, so hpcloud/tail started at offset 0 and **re-read the whole file on every repman start/restart** — re-parsing/re-fingerprinting all history and re-flooding the in-memory buffer. Now seeks to `io.SeekEnd` (standard `tail -f`; the buffer is a rolling recent window).

### 3. Self-heal a runaway collected log (#1665)
A client can already be sitting on a multi-GB file left by the pre-fix re-stream. On tailer open, if a collected DB log exceeds a hard ceiling (`selfHealDBLogCeilingMB`, default 1 GB, **independent of the `db-log-rotate` flag**), repman moves it aside, recreates an empty active file, gzips the moved-aside copy in the background (keeps a `.gz` sample), and raises **WARN0208**. Safe only because #1661/#1663 stop the re-inflation, so the bulk is confirmed garbage — and a compressed sample + state are still preserved (not a silent truncate).

### 4. gzip compression generalized to every repman log (#1663)
Compression was set on `carbonapi.log` only. Wired `Compress: true` into the shared `NewRotateFileHook` and `NewRotateWriter`, so **every** repman log writer (main `replication-manager.log`, security/workload/schema/monitor, both cluster logs, and all four fetched DB logs) gzips rotated backups.

## Explicitly out of scope (deferred)

`DBLogRotate` / `db-log-rotate` **flag semantics** — its correct meaning is DB-**node**-side rotation behavior (default off = don't touch the client's own DB logs), not gating repman-side rotation. Re-scoping it touches the runtime toggle API, GUI, and existing tests, and is a separate decision. **None of the fixes here depend on it.**

## Testing

- `go build -tags server ./...` clean.
- New unit tests: `TestMaybeSelfHealOversizedDBLog`, `TestMaybeSelfHealOversizedDBLog_LeavesSmallFileAlone`.
- `bash -n share/scripts/dbjobs_new.sh` clean.
- Existing DB-log/tailer/rotation test suite green.
- Targeted for the OpenSVC topology-matrix test loop before merge (dbjob change propagates to all DBs).

## Cleanup note

The already-inflated file on dbaas-fr-2 is left in place (no live truncation performed — investigation stayed read-only); once this ships, the self-heal reclaims it automatically on next tailer open.

### DIFF
--- a/cluster/srv.go
+++ b/cluster/srv.go
@@ -16,6 +16,7 @@ import (
 	"errors"
 	"fmt"
 	"hash/crc64"
+	"io"
 	"net/http"
 	"os"
 	"os/exec"
@@ -573,7 +574,15 @@ func (server *ServerMonitor) NewLogTailer(logtype string) (*tail.Tail, error) {
 		}
 	}
 
-	return tail.TailFile(logfile, tail.Config{Follow: true, ReOpen: true})
+	// Seek to end on open: follow only lines appended after startup. Without a
+	// Location, hpcloud/tail starts at offset 0 and re-reads the ENTIRE file on
+	// every repman start/restart -- re-parsing and re-fingerprinting every
+	// historical line and re-flooding the in-memory buffer. On a large collected
+	// log that is a pure-waste CPU spike. Standard tail -f semantics: the buffer
+	// is a rolling recent window, so lines appended while repman was down are
+	// intentionally skipped rather than re-processed.
+	return tail.TailFile(logfile, tail.Config{Follow: true, ReOpen: true,
+		Location: &tail.SeekInfo{Offset: 0, Whence: io.SeekEnd}})
 }
 
 func (server *ServerMonitor) Ping(wg *sync.WaitGroup) {
--- a/cluster/srv.go
+++ b/cluster/srv.go
@@ -15,6 +15,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"compress/gzip"
 	"hash/crc64"
 	"io"
 	"net/http"
@@ -567,6 +568,10 @@ func (server *ServerMonitor) NewLogTailer(logtype string) (*tail.Tail, error) {
 		misc.RemoveOldLogFiles(logDir, fmt.Sprintf("%s_", logName), cluster.Conf.DBLogRotateMaxAge, "20060102_150405")
 	}
 
+	// Self-heal a runaway collected log (broken-deployment artifact) before we
+	// attach the tailer, so we don't reopen/follow a multi-GB file.
+	server.maybeSelfHealOversizedDBLog(logfile)
+
 	if _, err := os.Stat(logfile); os.IsNotExist(err) {
 		nofile, ferr := os.OpenFile(logfile, os.O_WRONLY|os.O_CREATE, 0600)
 		if ferr == nil {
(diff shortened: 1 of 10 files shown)


svaroqui:
@claude please do a thorough code review of this PR.

Scope — makes repman's own log handling *perpetual* (no repman-written file grows unbounded):
- `share/scripts/dbjobs_new.sh` `dblogfile()`: persist the stream offset (was read-but-never-written) + fix `$TMPDIR`→`$TMP_DIR`. Verify the offset resume logic is correct across rotation/truncation of the DB-side source and can't skip or duplicate lines.
- `cluster/srv.go` `NewLogTailer`: seek collected logs to EOF (stop re-parsing whole file on restart); `maybeSelfHealOversizedDBLog` (rename-aside + background gzip + WARN0208) — check the race/ordering around recreating the empty file before the tailer attaches, the SST-in-flight guard, and that the goroutine can't leak.
- `cluster/srv.go`: make rotated DB-log history pruning unconditional (no longer gated by `db-log-rotate`).
- `utils/s18log/rotatefilehook.go`: `Compress:true` on both shared writers; `RemoveOldLogFilesWithExt`.
- `cluster/srv_get.go`: PFS snapshot retention (`monitoring-pfs-snapshot-retention-days`, default 2) — confirm it prunes only the hourly series and never the Schema-Graph-served files incorrectly.

Please focus on correctness (data races, goroutine/file-handle leaks, error paths), and whether the retention/self-heal can ever delete a file that's still being written or that a reader/GUI depends on. Deliberately out of scope: the `db-log-rotate` flag semantics (#1667) and the event log (#1669).

PR #1670, opened 2026-08-06
Fix #1666 code-review findings: self-heal data-loss race + PFS prune skipped on idle servers

svaroqui:
Follow-up to #1666, addressing the code-review findings on that PR.

## Fixes

**Finding 1 (HIGH — silent data loss)** — `maybeSelfHealOversizedDBLog` guarded only on `IsFileOpenForSSTReceive`, missing the long-lived cached lumberjack writer that `GetSlowLogTable` / the SST rotate path hold open via `getDBLogRotatingWriter`. Renaming the file out from under that writer sent its later writes to the renamed inode, which `gzipFileAndRemove` then compressed + deleted → lost slow-query data. Now self-heal also **skips when a cached writer holds the path** (that writer already bounds the file). New regression test `TestMaybeSelfHealOversizedDBLog_SkipsWhenCachedWriterHoldsFile`.

**Finding 2 (HIGH — unbounded growth)** — the PFS snapshot retention prune sat *after* the "nothing to flush" early-return, so idle/erroring servers never reached it and `log_pfs_queries_*.jsonl` grew unbounded — the exact bug class #1666 was written to fix. Moved the prune **ahead of the early-return** so it runs every cycle.

## Deliberately deferred (tracked separately)

**Finding 3 (Low/Med — duplicate gzip)** — `gzipFileAndRemove` reinvents what the rotating writer already does. The correct fix is to make the repman-side collected-log writer **always** go through the rotating writer (lumberjack), so a runaway file is rotated+gzipped+pruned by the one mechanism — which would delete `maybeSelfHealOversizedDBLog` + `gzipFileAndRemove` entirely and also remove the Finding-1 race by construction. That's a larger refactor with a test cascade, intentionally **not** rushed right before an official release. Tracking issue to follow.

## Testing
- `go build -tags server ./...` clean.
- `go test ./cluster -run 'SelfHeal|PFS|Flush|Prune|Tailer'` green, incl. the new guard test.

### DIFF
--- a/cluster/srv.go
+++ b/cluster/srv.go
@@ -621,6 +621,18 @@ func (server *ServerMonitor) maybeSelfHealOversizedDBLog(logfile string) {
 	if cluster.IsFileOpenForSSTReceive(logfile) {
 		return
 	}
+	// Don't race a long-lived cached rotating writer either. GetSlowLogTable and
+	// the SST rotate path borrow a shared *lumberjack.Logger for this exact path
+	// (getDBLogRotatingWriter) that holds the file open by fd. If we renamed the
+	// file out from under it, its subsequent writes would go to the renamed inode
+	// -- which gzipFileAndRemove then compresses and deletes -- silently losing
+	// data. A cached writer also already bounds the file, so defer to it.
+	cluster.dbLogWriterMutex.Lock()
+	_, hasCachedWriter := cluster.dbLogWriters[logfile]
+	cluster.dbLogWriterMutex.Unlock()
+	if hasCachedWriter {
+		return
+	}
 
 	ts := time.Now().Format("20060102_150405")
 	backup := strings.TrimSuffix(logfile, ".log") + "_oversize_" + ts + ".log"
--- a/cluster/srv_get.go
+++ b/cluster/srv_get.go
@@ -640,14 +640,6 @@ type PFSSnapshotEntry struct {
 func (server *ServerMonitor) FlushPFSSnapshotToLog() {
 	cluster := server.ClusterGroup
 
-	// We need an up-to-date fetch to get sample queries before the TRUNCATE.
-	pfsq, _, err := dbhelper.GetQueries(server.Conn, server.DBVersion)
-	if err != nil || len(pfsq) == 0 {
-		cluster.LogModulePrintf(cluster.Conf.Verbose, config.ConstLogModTask, config.LvlDbg,
-			"PFS snapshot: nothing to flush on %s", server.URL)
-		return
-	}
-
 	os.MkdirAll(server.Datadir+"/log", 0755)
 
 	// Unconditional repman-side housekeeping: prune hourly PFS snapshot files
@@ -655,11 +647,24 @@ func (server *ServerMonitor) FlushPFSSnapshotToLog() {
 	// per-hour files ARE a feature (the Schema Graph time-series browses them
 	// via /api .../pfs-snapshots), so we keep a rolling window rather than
 	// merging them; the window is monitoring-pfs-snapshot-retention-days.
+	//
+	// This runs BEFORE the "nothing to flush" early-return below so it fires on
+	// EVERY cycle -- an idle server, or one whose PFS fetch is temporarily
+	// disabled/erroring, must still have its old snapshots pruned, otherwise the
+	// series grows unbounded exactly like the bug this PR was written to fix.
 	retentionDays := cluster.Conf.MonitorPFSSnapshotRetentionDays
 	if retentionDays <= 0 {
 		retentionDays = 2
 	}
 	misc.RemoveOldLogFilesWithExt(server.Datadir+"/log", "log_pfs_queries_", ".jsonl", retentionDays, "20060102_15")
+
+	// We need an up-to-date fetch to get sample queries before the TRUNCATE.
+	pfsq, _, err := dbhelper.GetQueries(server.Conn, server.DBVersion)
+	if err != nil || len(pfsq) == 0 {
+		cluster.LogModulePrintf(cluster.Conf.Verbose, config.ConstLogModTask, config.LvlDbg,
+			"PFS snapshot: nothing to flush on %s", server.URL)
+		return
+	}
 
 	now := time.Now()
 	filename := server.Datadir + "/log/log_pfs_queries_" + now.Format("20060102_15") + ".jsonl"
--- a/cluster/srv_selfheal_test.go
+++ b/cluster/srv_selfheal_test.go
@@ -68,6 +68,47 @@ func TestMaybeSelfHealOversizedDBLog(t *testing.T) {
 	}
 }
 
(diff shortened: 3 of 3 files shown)


tanji:
## Automated review (recall-biased, high effort)

Reviewed the diff (`cluster/srv.go`, `cluster/srv_get.go`, `cluster/srv_selfheal_test.go`) against the full PR context (fix for #1666 review findings). 8 finder angles + 1-vote verification. Findings below, most severe first.

### 1. `hasCachedWriter` check treats an idle cached writer the same as an active one — can permanently disable self-heal for a path (CONFIRMED)
**`cluster/srv.go:630-635`**

The new guard is:
```go
cluster.dbLogWriterMutex.Lock()
_, hasCachedWriter := cluster.dbLogWriters[logfile]
cluster.dbLogWriterMutex.Unlock()
if hasCachedWriter {
    return
}
```
This only checks map-key *presence*, not `entry.borrowers > 0` or `!entry.stale`. Per `releaseDBLogWriterFunc` in `cluster/srv_job_logs.go:364-383`, a non-stale entry is **never removed from the map once its last borrower releases it** (`if entry.borrowers > 0 || !entry.stale { return }` — deletion only happens when both borrowers==0 *and* stale==true). The only eviction paths (`CloseAllDBLogWriters`, `pruneStaleDBLogWriters`) fire solely on `db-log-rotate` being disabled, cluster teardown, or a reload/path change — never on idle-timeout.

Net effect: the *first* time anything acquires a writer for a given log path (e.g. one on-demand `GetSlowLogTable` export), that path is excluded from self-heal for the remaining life of the process, even after the writer goes idle (0 borrowers) and even if the file keeps growing through some other means. This defeats the self-heal feature it's meant to protect, for any path that was ever touched once. Given the writer's own doc comments frame a cached-but-idle entry as legitimately still "owning" the fd for reuse, this may be intentional — but if so, it trades the data-loss race for a correctness regression (oversized files never reclaimed), and that tradeoff isn't called out in the PR description or tested.

### 2. `migrateDBLogsToBackupStorage` has the identical unguarded rename race this PR just fixed elsewhere (CONFIRMED)
**`cluster/srv_job_logs.go:466-479`**

This function (rename/copy legacy DB log files to backup-backed storage on a `db-log-on-backup-storage` flip) checks only `cluster.IsFileOpenForSSTReceive(src)` before calling `renameOrCopyFile(src, dst)` — it never checks `cluster.dbLogWriters`. It's reachable from `RestartDBLogTailers` (API-triggered flag flip) concurrently with `GetSlowLogTable`'s cached writer (also acquired via `getDBLogRotatingWriter` on a per-server monitoring goroutine, `cluster.go:1284`) actively appending to the same legacy-dir path. Nothing serializes the two, so the exact "rename out from under a live cached writer → silently lost writes" scenario this PR fixes in `maybeSelfHealOversizedDBLog` is still open here. Given the PR's stated purpose is closing this race class, this sibling gap is worth addressing in the same pass (or explicitly tracked, the way Finding 3 is).

### 3. New regression test locks in the map-presence bug rather than proving the intended "don't race an active writer" semantics (CONFIRMED)
**`cluster/srv_selfheal_test.go:96-116`**

`TestMaybeSelfHealOversizedDBLog_SkipsWhenCachedWriterHoldsFile` seeds `map[string]*dbLogWriterEntry{logfile: {}}` — a zero-value entry (`writer: nil, borrowers: 0, stale: false`) that `getDBLogRotatingWriter` never produces in practice (real entries always have a non-nil writer and `borrowers >= 1` while genuinely held). If the guard were later tightened to `entry.borrowers > 0`, this test's seeded entry (`borrowers: 0`) would fail the check and the test's "must be skipped" assertions would break — meaning the test verifies map-presence, not "a writer is actually holding this file open," and would not catch a fix/regression of finding #1 either way.

### 4. TOCTOU gap between the cached-writer check and `os.Rename` (PLAUSIBLE)
**`cluster/srv.go:630-640`**

`dbLogWriterMutex` is unlocked immediately after the map read; `os.Rename` happens several lines later with no lock held across the gap. A concurrent `getDBLogRotatingWriter` call (e.g. from `GetSlowLogTable`, launched as an independent per-server goroutine via `cluster.trackTickGoroutine`) can create and cache a fresh writer for the same path in that window, after which the rename proceeds anyway and races the new writer — narrowing but not eliminating the class of bug this PR fixes. Note the pre-existing `IsFileOpenForSSTReceive` check just above has the same shape, so this isn't a new pattern, but it means the fix is not airtight under concurrency.

### 5. PFS snapshot prune still never runs when PFS monitoring is disabled, despite the PR's "runs on every cycle" framing (PLAUSIBLE)
**`cluster/srv_get.go:640-664`**

`FlushPFSSnapshotToLog` (and its now-reordered prune) is only reachable through `GetPFSQueries`, which itself early-returns before ever calling it when `!cluster.Conf.MonitorPFS || !server.HavePFS` (`cluster/srv_get.go:522-528`). If PFS monitoring is later disabled on a server that previously accumulated `log_pfs_queries_*.jsonl` files, those files are now orphaned and never pruned by anything — the comment's claim that pruning "fires on EVERY cycle" and "must still have its old snapshots pruned" is true only for the "GetQueries errors/empty" case the PR targeted, not the "PFS monitoring off entirely" case, which remains unbounded-growth-prone.

---

Lower-priority / cleanup notes (not blocking):
- The new lock/lookup/unlock block in `maybeSelfHealOversizedDBLog` duplicates a pattern already inlined five other times across `cluster/srv_job_logs.go`; a small `cluster.hasCachedDBLogWriter(logfile string) bool` helper would centralize it and pair naturally with `IsFileOpenForSSTReceive`.
- Altitude: this is now the third independent "is this file busy" guard bolted onto `maybeSelfHealOversizedDBLog` (SST-receive check, now the writer-cache check). The PR body itself acknowledges (Finding 3) that the durable fix is routing everything through the rotating writer so a second, independent size-poll-and-rename mechanism doesn't need to keep growing bespoke guards for every new writer type.

svaroqui:
Re-review (high-effort agent) found 3 defects — **this patch should not be merged**:
1. **TOCTOU data-loss race NOT closed** — the cached-writer guard releases the mutex before `os.Rename`, so a concurrent `getDBLogRotatingWriter` can still open the file between check and rename → the rename→gzip→delete slow-query data-loss race remains.
2. Guard **permanently disables self-heal** whenever any writer is cached — breaks reclamation if `db-log-rotate-max-size` > the 1GB self-heal ceiling.
3. PFS retention prune still only runs when `monitoring-pfs-queries` is on (`FlushPFSSnapshotToLog` isn't called when PFS is off) → unbounded growth persists.

These can't be cleanly patched here. Correct fix = **#1671** (always-on rotating writer; delete `maybeSelfHealOversizedDBLog` + `gzipFileAndRemove`; move PFS retention to an always-run site) — removes all three by construction. **Superseded by #1671; closing.**

Client safety is unaffected: the runaway-log fix (#1666 — offset persistence + tail-from-EOF) is already merged and verified live; this PR was only the secondary reclaim path.

Issue #1675, opened 2026-08-07
Unbounded graphite metrics queue → OOM when carbon sink cannot keep up (memory leak)


svaroqui:
### Summary

`cluster.(*ClusterGraphite).metrics` is an **unbounded in-memory queue**. It is appended to on every monitoring tick and drains **only when a send fully succeeds**; on any send failure the whole batch is re-queued. With no size cap, any sustained send failure makes the queue grow at the full metric-production rate → **RSS grows without bound → OOM**.

This is the second hot path reported in #1633 (after the InnoDB-status leak was fixed in v3.1.38): the reporter's `inuse_space` heap on v3.1.38 showed
```
793.55MB 51.89%  fmt.Sprintf
560.27MB 36.63%  cluster.(*ClusterGraphite).AddMetrics
         (under FetchDatabaseStats -> GetDatabaseMetrics)
```
Both frames are the **same retained slice** — the `[]graphite.Metric` queue and the metric name/value strings held inside it.

### Why it triggers even with an *embedded* carbon sink

"Embedded carbon process is running" does **not** mean "the send succeeds." Each flush still crosses a real TCP socket to `graphite-carbon-host:graphite-carbon-port` (default `127.0.0.1:2003`) with a 1s deadline:

1. **Reconnect-per-flush / connection refused.** `SendGraphiteMetrics` calls `Connect()` every flush, and `graphite.Connect()` unconditionally closes and re-dials (`net.DialTimeout`). If the carbon listener isn't accepting on the configured port (not started, port mismatch, `graphite-metrics` enabled while the listener isn't) → connection refused every flush → error → requeue.
2. **1s write-deadline timeout under backpressure.** `sendMetrics` sets `conn.SetWriteDeadline(now + Timeout)` with `Timeout` defaulting to **1s**, then a single `conn.Write`. If the embedded carbon ingests slower than repman produces (kernel socket buffer fills), the write times out → error → requeue. At scale (N servers × M metrics/tick) an embedded carbon can absolutely fall behind.

Either path: **error → `requeue()` → the uncapped queue grows every tick.**

### Relationship to `b4e744eaf` (`fix(graphite): stop flush stalls from retaining metrics`)

[...]

PR #1677, opened 2026-08-07
fix(graphite): bound per-cluster metrics queue

caffeinated92:
This PR is related to #1675 

## What changed
- add a per-cluster graphite-metrics-queue-limit hard cap with a safe default fallback
- bound both AddMetrics and requeue with a drop-oldest policy that rebuilds a fresh backing slice
- track sustained drops with DroppedMetricsTotal and WARN0192
- reuse healthy Graphite connections and reconnect only after deadline/write failure
- add regtest coverage and implementation documentation for the queue-bound behavior

## Validation
- go test ./cluster -run TestClusterGraphite|TestWARN0192
- go test ./graphite
- go test ./server -run TestSetClusterSetting_GraphiteMetricsQueueLimit|TestSetClusterSetting_GraphiteMetricsQueueLimit_IndependentPerCluster
- go test ./regtest


caffeinated92:
@claude this is my local review:

Re-review: PR #1677 (memory-leak branch), current state

  All fixes from this session are now committed (auto-committed/pushed as 7e84564e0 —
  flagging again since that pushed to origin/memory-leak without an explicit git push from
  me). Full re-verification:

  Build & tests — clean.
  - go build -tags server ./... — OK.
  - go test -race ./cluster/... ./graphite/... ./server/... (Graphite-relevant subset) — all
  pass, including the new
  TestClusterGraphite_*/TestWARN0192_*/TestSetClusterSetting_GraphiteMetricsQueueLimit*
  tests, with no data races.
  - The broader -race run surfaced failures in unrelated packages (utils/backupmgr Restic
  races, graphite/expr, graphite/logging, graphite/persister, graphite/receiver,
  cluster/logplugin audit-drift). I checked each against a clean develop worktree — same
  failures reproduce there, and restic.go is byte-identical between the two branches.
  Pre-existing, unrelated to this PR, not something to hold #1677 on.

  All three previously-found bugs are fixed and verified:
  1. Data race on cg.gc — fixed with the dedicated gcLock, scoped narrowly (never held
  across SendMetrics/across AddMetrics). Re-ran my synthetic race reproduction against the
  fix: clean.
  2. NumberInput 0-value staleness — fixed (value !== undefined && value !== null). Checked
  it doesn't regress any of the ~10 other call sites (backupKeep*, PFS settings,
  RefreshCounter, etc.) — if anything it fixes latent bugs in several of them too, since
  they share the component.
  3. Dead ternary in GraphSettings.jsx — removed, now matches the plain-passthrough pattern
  used elsewhere (RepConfigSettings.jsx).

  Still open, unchanged from before, not blocking:
  - T8: technical doc (GRAPHITE_SCOPE.md) is solid; the docs.signal18.io user-doc side can't
  be verified from this repo.
  - api_cluster.go's strconv.Atoi error-swallow: left as-is, consistent with 15+ sibling
  cases in the same switch — fixing only this one would be the T19-inconsistent move.
  - Stale-connection-on-write edge in graphite.go: accepted trade-off, self-healing via
  requeue, strictly better than the old always-redial behavior.
  - New finding from the SetGraphiteConnection discussion: graphite-carbon-host/-port API
  changes don't appear to reach a running cluster's Conf at all (setClusterSetting has no
  case for them, error silently discarded in the fan-out) — this is a pre-existing, separate
  config-propagation gap, not caused by #1677. Worth its own issue rather than folding into
  this one.

  Bottom line: #1677 is now in good shape against DEVELOPMENT_LAWS.md — T18's
  bounding/tracking is correctly implemented and verified race-free, the UI bugs you flagged
  are fixed, and the remaining items are either pre-existing/out-of-scope or intentional
  trade-offs I've explained. Good to merge from a laws-compliance standpoint, modulo the T8
  external-doc confirmation.

Issue #1633, opened 2026-07-29
[Bug]: ## Memory leak in GetEngineInnoDBStatus / GetEngineInnoDBVariables (monitoring-innodb-status)


windyHe2020:
### Component

Server (monitoring/orchestration)

### Bug Description

Enabling `monitoring-innodb-status` causes unbounded memory growth over time.
pprof heap profiling shows `database/sql.convertAssignRows`, reached via
`utils/dbhelper.GetEngineInnoDBStatus` → `GetEngineInnoDBVariables` →
`cluster.(*ServerMonitor).FetchDatabaseStats` → `Ping` → `Refresh`,
retaining the large majority of inuse heap space after ~11 days of uptime.

### Steps to Reproduce

- replication-manager version: v3.1.30 (confirmed still present through v3.1.37 by code inspection; commit history shows no changes to `GetEngineInnoDBStatus`/`GetEngineInnoDBVariables` since their introduction other than the `utils/dbhelper` file-split refactor in `02e39ed`)
- Monitored DB: Percona/MySQL 8.0.32, GTID replication
- OS: Linux (Rocky/CentOS-based), self-built binary
- Config: `monitoring-innodb-status = true`

1. Enable `monitoring-innodb-status = true` on a cluster monitoring an active MySQL 8.0.32 instance.
2. Let repman run for several days under normal monitoring load (default polling interval).
3. Observe process RSS growth (~500MB+/day in our environment).
4. Capture heap profile: `curl http://127.0.0.1:10001/debug/pprof/heap > heap.out`
5. `go tool pprof -top heap.out`

### Expected Behavior

Normal memory growth

### pprof output (inuse_space, after ~11 days uptime)

Showing nodes accounting for 5113.27MB, 99.06% of 5161.66MB total
flat flat% sum% cum cum%
4182.68MB 81.03% 81.03% 4182.68MB 81.03% database/sql.convertAssignRows
572.04MB 11.08% 92.12% 572.04MB 11.08% fmt.Sprintf
358.55MB 6.95% 99.06% 358.55MB 6.95% cluster.(*ClusterGraphite).AddMetrics
0 0% 99.06% 4182.68MB 81.03% database/sql.(*Rows).Scan
0 0% 99.06% 4182.68MB 81.03% database/sql.(*Rows).scanLocked
0 0% 99.06% 933.09MB 18.08% cluster.(*ServerMonitor).FetchDatabaseStats
[...]

PR #1679, opened 2026-08-08
fix(graphite): bound the metrics send queue to stop the OOM (#1675)

### LINKED ISSUE #1675: Unbounded graphite metrics queue → OOM when carbon sink cannot keep up (memory leak)

svaroqui:
Closes #1675. Completes the second half of #1633 (the graphite hot path).

### Problem
`ClusterGraphite.metrics` is an **unbounded** in-memory send queue: it drains only on a fully successful send, requeues the whole batch on any failure, and producers append every tick. When the carbon sink can't keep up — unreachable, connection-refused, or a 1s write-deadline timeout (which an **embedded** carbon can hit under backpressure) — the queue grows at the full production rate until the monitor OOMs. This is the `fmt.Sprintf` + `ClusterGraphite.AddMetrics` retention in the #1633 profile that remained after the InnoDB fix.

### Prior art (why this shape)
The upstream client lib (`marpaia/graphite-golang`) is fire-and-forget; the component that buffers a slow sink in the graphite world is the **carbon relay**, which uses a **bounded `MAX_QUEUE_SIZE` + bounded retry and drops under sustained pressure**. repman embeds its own carbon, so it plays that relay role and the in-process queue must be bounded the same way.

### Fix
Two commits:
1. `stop flush stalls from retaining metrics` (cherry-picked from `memory-leak`, @caffeinated92) — removes the lock-across-network stall, adds a 1s write deadline.
2. `bound the metrics send queue` — the missing cap:
   - `graphite-metrics-max-queue-size` config (named after carbon's `MAX_QUEUE_SIZE`; default 100000; `0` ⇒ default, never unbounded).
   - `capMetricsLocked` drops the **oldest** metrics past the cap in both `AddMetrics` and `requeue`, copying the newest into a fresh backing array so dropped metrics/strings are actually freed.
   - drops tracked (`DroppedMetrics`) and surfaced as **WARN0209**, held by the `%5` `PreserveState` block so the alert doesn't flap.

Commit 1 alone is necessary but not sufficient — its `requeue()` still re-appends with no cap, converting a stall into unbounded growth. Commit 2 bounds it.

### Tests
Unit tests cover the cap in `AddMetrics` and `requeue` (drop-oldest, keep-newest, drop count). A real multi-product regtest (bounded RSS with graphite-metrics on and no reachable carbon) is a follow-up.

### Law
Enforces the perpetual-monitoring invariant (functional laws F3+F4 / DEVELOPMENT_LAWS T18): a telemetry buffer must be bounded — a monitor that OOMs on a failing sink stops monitoring.


svaroqui:
Superseded by PR #1677 (already merged to develop), which implements the same queue bound more completely — `boundMetrics`/`GraphiteMetricsQueueLimit`/`DroppedMetricsTotal` plus a regtest, connection guard, and docs. My branch duplicated in-flight work on the same issue; closing in favor of what's on develop. The distinct passive-monitor gate lives on in #1681.

PR #1705, opened 2026-08-14
fix(cluster): harden logical mysqldump reseed flow

caffeinated92:
## Summary
- unify file-based mysqldump logical reseed with the two-phase system replay path used by direct reseed
- centralize logical restore-user and split-user resolution so custom backup paths and stale metadata cannot select the wrong replay behavior
- dispatch WARN0075 logical reseeds from the open state with an atomic guard so scheduler mode changes cannot miss or duplicate launches
- remove duplicate logical reseed preflight logs and document the delivered behavior

## What this fixes
- prevents file-based mysqldump reseeds from replaying the wrong user/system SQL source
- fixes mysql.users.sql.gz fallback handling, including missing-sidecar no-op behavior when inline system content is absent
- keeps backup-restore-mysql-user behavior consistent for logical reseed and flashback even when metadata is stale, unrelated, or overridden
- prevents WARN0075 logical reseeds from stalling, relaunching after completion, or dispatching concurrently around scheduler mode flips
- keeps logical reseed preflight reporting accurate without duplicate info logs

## Validation
- `go test ./cluster -run "Test(StateProcessing_WARN0075StaleTriggerIsNoop|StateProcessing_WARN0075SurvivesModeFlipMidReseed|LaunchLogicalReseed_ConcurrentCallsDispatchOnce|ReseedMysqldump.*|ClassifyReseedMysqldump.*|LogicalReseed.*|ResolveLogicalReseed.*)"`
- `go test ./cluster/...`
- focused logical reseed coverage passed
- existing `cluster/logplugin` audit drift failures reproduce on both this branch and `origin/develop`:
  - `TestAuditDrift_NewTemplateDetected`
  - `TestAuditDrift_CustomWindowViaConfig`
  - `TestAuditDrift_MaxNewTemplatesList`

## Follow-up
- T13 Docker/regtest coverage for file-based logical reseed and flashback remains to be added for full matrix validation


caffeinated92:
@claude review my PR.

All three code review items are fixed and verified:

  1. Sidecar reader leak — ReadMysqldumpUser now returns io.ReadCloser (a small
  gzipFileReadCloser wrapper closing both the gzip stream and the underlying file); gzfile
  is also closed if gzip.NewReaderN fails; classifyReseedMysqldumpUserSidecar now defers the
  close.
  2. Pointer aliasing — replaced splitUserOverridePtr = &splitUser with an explicit stable
  copy (overrideVal := splitUser; splitUserOverridePtr = &overrideVal), so it no longer
  relies on Go's evaluation-order guarantee to be correct.
  3. Dead SplitUser field — removed from logicalReseedUserRestoreAssessment (struct def, the
  one assignment site, and the test literal that referenced it); confirmed via grep it had
  no other production or test readers.
  4. T13 gap — left as-is: the branch's own status doc already discloses it honestly as a
  mandatory-but-deferred follow-up requiring real Docker version-matrix fixtures, which
  isn't something to bolt on as a quick regtest in this pass.

caffeinated92:
@claude review my PR once more

PR #1711, opened 2026-08-19
fix(cluster): protect staging standalone recovery from read-only

caffeinated92:
## Summary
- identify the active staging standalone from `staging-server-host` during recovery checks
- skip the automatic read-only transition for that staging standalone only
- add unit and targeted regtest coverage for the staging recovery path

## Testing
- go test ./cluster/... ./regtest/... ./server/...


caffeinated92:
@claude review my PR

PR #1723, opened 2026-08-20
fix(deps): pin cilium/ebpf to v0.20.0

### LINKED ISSUE #1722: Build break: cilium/ebpf v0.22.0 is incompatible with containerd/cgroups v1.0.1
## Summary

At revision `0175926ad`, the module graph resolves both:

- `github.com/cilium/ebpf v0.22.0`
- `github.com/containerd/cgroups v1.0.1`

These versions do not compile together.

`containerd/cgroups v1.0.1` still uses the old `asm.Instruction.Sym(...)` API in `v2/devicefilter.go`, while `cilium/ebpf` removed that API in favor of `WithSymbol(...)`.

This produces a build failure similar to:

```text
asm.Instruction has no field or method Sym
```

## Root cause

The `cilium/ebpf` bump came from dependabot commit:

- `c316e0041` — `chore(deps): bump github.com/cilium/ebpf from 0.11.0 to 0.22.0`

The old `containerd/cgroups v1.0.1` is still brought in transitively via `github.com/opensvc/om3/v3`.

Important scope note:

- **`om3` itself is out of scope for this fix**
- this issue is about how `replication-manager` resolves the transitive dependency set

## Reproduction

On the affected revision:

```bash
make cli
```

and/or inspect the module graph:

```bash
go list -m github.com/cilium/ebpf github.com/containerd/cgroups
go mod graph | rg "github.com/cilium/ebpf|github.com/containerd/cgroups"
```

## Short-term fix (preferred immediate unblock)

Pin `github.com/cilium/ebpf` to `v0.20.0` in the root module.

Why `v0.20.0`:

- it is still **before** the `Sym` removal in `cilium/ebpf v0.21.0`
- it still supports `Instruction.Sym(...)` for `containerd/cgroups v1.0.1`
- it also supports `Instruction.WithSymbol(...)`, which is used by `containerd/cgroups/v3`
- it keeps the fix inside `replication-manager` scope without changing `om3`

This is a narrower compatibility pin than a full revert back to `v0.11.0`.

## Alternative short-term fix

A full revert of the dependabot bump (`c316e0041`) should also restore compatibility, but pinning to `v0.20.0` is the more targeted compromise currently being staged.

## Longer-term fix

Keep newer ebpf, but patch or override the transitive `github.com/containerd/cgroups` dependency at the root module level so it no longer uses `.Sym(...)`.

This keeps the fix inside `replication-manager` scope without requiring an `om3` upgrade.

## Acceptance criteria

- `make cli` builds successfully again
- no `asm.Instruction ... Sym` compile errors remain
- no `om3` upgrade is required for the immediate fix
- the selected short-term dependency set is compatible with both `containerd/cgroups v1.0.1` and `containerd/cgroups/v3`

caffeinated92:
## Summary
- pin `github.com/cilium/ebpf` from `v0.22.0` to `v0.20.0`
- restore compatibility with transitive `github.com/containerd/cgroups v1.0.1`
- keep the fix scoped to `replication-manager` without changing `om3`

## Why
`containerd/cgroups v1.0.1` still uses `asm.Instruction.Sym(...)`, which is removed in `cilium/ebpf v0.21.0+`.

`v0.20.0` is the narrow compatibility pin because it still supports both:
- `Instruction.Sym(...)`
- `Instruction.WithSymbol(...)`

That keeps compatibility with both `containerd/cgroups v1.0.1` and `containerd/cgroups/v3`.

Closes #1722

## Validation
- `make cli` ✅
- `go test ./opensvc/...` ❌ pre-existing failure, reproduced on `origin/develop` in the same `TestCreateV2_*` cases

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -136,7 +136,7 @@ require (
 	github.com/aws/aws-sdk-go-v2/service/internal/presigned-url v1.13.21 // indirect
 	github.com/aws/aws-sdk-go-v2/service/internal/s3shared v1.19.21 // indirect
 	github.com/cenkalti/backoff/v4 v4.3.0 // indirect
-	github.com/cilium/ebpf v0.22.0 // indirect
+	github.com/cilium/ebpf v0.20.0 // indirect
 	github.com/clipperhouse/uax29/v2 v2.2.0 // indirect
 	github.com/cloudflare/circl v1.6.3 // indirect
 	github.com/containerd/cgroups v1.0.1 // indirect
--- a/go.sum
+++ b/go.sum
@@ -145,8 +145,6 @@ github.com/cespare/xxhash/v2 v2.3.0/go.mod h1:VGX0DQ3Q6kWi7AoAeZDth3/j3BFtOZR5XL
 github.com/cilium/ebpf v0.4.0/go.mod h1:4tRaxcgiL706VnOzHOdBlY8IEAIdxINsQBcU4xJJXRs=
 github.com/cilium/ebpf v0.20.0 h1:atwWj9d3NffHyPZzVlx3hmw1on5CLe9eljR8VuHTwhM=
 github.com/cilium/ebpf v0.20.0/go.mod h1:pzLjFymM+uZPLk/IXZUL63xdx5VXEo+enTzxkZXdycw=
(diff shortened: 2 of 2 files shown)


caffeinated92:
@claude review my PR

PR #1725, opened 2026-08-20
feat(logs): add log filters and bounded history (#1714)

### LINKED ISSUE #1714: [Feature]: GUI log viewer — filter by level (and module/time)
## Component
Dashboard (GUI) — log viewer

## Summary
Add **level filtering** to the GUI log viewer so an operator can browse repman logs and narrow to ERR / WARN / INFO / DBG (and ideally by module and time), instead of reading an unfiltered stream. Today the only way to find, say, all warnings or a switchover event is to SSH to the host and `grep` the raw log file — slow and error-prone.

## Motivation
Investigating recent events (e.g. "when did cluster X switch over?", "show me only errors in the last hour") currently requires host access + `grep` over multi-hundred-MB log files. The structured data to do this in the GUI already exists; it just isn't exposed as a filter.

## Current state (already in place)
- **Data model already carries level**: `s18log.HttpMessage { Group, Level, Timestamp, Text, Module }` — every buffered log line has a `Level` and `Module`.
- **API endpoints already serve the ring buffer**: global logs via `handlerMuxGlobalLogs` (`/api/logs`) and the per-cluster log endpoint.
- **GUI Logs component already exists** (`share/dashboard_react/src/Pages/Dashboard/components/Logs/index.jsx`) with a **text search** box (`search` → `text.includes`).

What is missing: no **level** filter, no **module** filter — the buffer is shown whole (minus free-text search).

## Proposed scope
1. **Level filter (primary):** a multi-select / segmented control (ERR, WARN, INFO, DBG, TRACE) in the Logs component. Since `Level` is already in the payload, filter client-side over the existing buffer to start — no backend change needed for v1.
2. **Module filter (secondary):** dropdown over the `Module` field (general, git, task, proxy, …) so a user can scope to one subsystem.
3. **Keep the existing text search**, combined (AND) with level/module.
4. **Nice-to-have (later):** server-side query params (`?level=`, `?module=`, `?since=`) on the log endpoint so filtering works beyond the in-memory ring buffer, and a time-range picker. If/when the API grows these params, keep the GUI as the surface (no API-only filtering — T6).

## Acceptance
- In the dashboard log view, selecting "WARN+ERR" shows only warning/error lines; combining with a text term further narrows.
- Filtering is instant on the already-loaded buffer.
- No regression to the existing search box or live-tail behavior.

## Notes
- Level values in the buffer follow the logrus strings (`error`/`warning`/`info`/`debug`/`trace`); map the UI labels to those.
- Per the laws: GUI-first is fine here; if a server-side filter API is added it ships with this GUI surface (T6), plus user + implementation docs (T8).

caffeinated92:
## Summary
- add bounded on-disk log history scanning for global and cluster general/task logs
- wire dashboard log filters, history loading, and time-range browsing into the existing log viewer
- add tests and implementation documentation for log history parsing, bounds, and merge behavior

## Testing
- go test ./utils/s18log ./server
- npm run build

Closes #1714


caffeinated92:
@claude review my PR changes

caffeinated92:
@claude review my PR once again independently

caffeinated92:
@claude review my PR changes

PR #1727, opened 2026-08-21
feat(haproxy): self-heal HAProxy 3.4 runtime dynamic backends

caffeinated92:
## Problem
replication-manager's HAProxy runtime integration could reconcile server state only when the expected read/write backends and rows already existed in the running HAProxy process. If `service_write`, `service_read`, or one of their expected rows was missing at runtime, repman could only log the failure and wait for a manual HAProxy reload.

## What this PR changes
This PR adds an **optional** HAProxy 3.4+ self-heal path for `haproxy-mode=runtimeapi` that can recreate missing runtime backends/servers without a reload.

### Main changes
- add `haproxy-runtime-dynamic-backends` to opt into the new behavior
- add HAProxy runtime API helpers for:
  - `add backend`
  - `publish backend`
  - `add server`
  - `enable server`
  - `enable health`
  - `del server`
  - `show servers state`
- extend `HaproxyProxy.Refresh()` to detect and self-heal:
  - missing write/read backends
  - missing `leader` write row
  - missing read rows
  - unhealthy read rows left behind by partial runtime operations
  - stale read rows for removed cluster members
- verify runtime operations instead of trusting `err == nil`, since HAProxy often reports command rejection as plain CLI text
- reject unsafe hostname-backed and stale-address runtime repairs instead of falsely reporting success
- support IPv4 and IPv6 runtime server creation/address verification

### Safety and compatibility
- feature is **off by default**
- self-heal runs only for:
  - HAProxy 3.4+
  - `haproxy-mode=runtimeapi`
- explicitly skipped for standby and staging paths
- existing generated proxy behavior is preserved by keeping the original `defaults` section unchanged and adding a separate trailing `defaults dyn_defaults` section used only by runtime-created backends via `add backend ... from dyn_defaults`
- OpenSVC's embedded HAProxy config was updated to match the same layout

## Tests and validation
- added focused regression coverage in `cluster/prx_haproxy_test.go` for:
  - backend publish verification
  - partial add/enable failures
  - retry behavior across refresh passes
  - stale-address detection
  - duplicate synthetic row prevention
  - same-pass master/read visibility
  - stale row deletion verification
  - IPv6 runtime rows
  - standby/staging gating
- technical implementation doc added at `doc/implementation/cluster/HAPROXY_DYNAMIC_BACKENDS.md`
- validated with:
  - `go test ./cluster/... -run Haproxy`
  - `go test ./cluster/...`
  - `go vet ./...`
  - live Docker checks against `haproxy:3.4-alpine`

## Scope notes
This PR keeps the existing fixed backend model (`service_write` / `service_read`) and does **not** attempt a broader shared-HAProxy or multi-cluster backend redesign.


caffeinated92:
@claude review my PR changes

caffeinated92:
@claude review my PR changes

PR #1738, opened 2026-08-26
fix(k8s): harden native provisioning correctness/lifecycle bugs

### LINKED ISSUE #1737: fix(k8s): harden native provisioning correctness/lifecycle bugs
## Summary

Pre-implementation audit of the native Kubernetes provisioning path (`cluster/prov_k8s.go`, `cluster/prov_k8s_db.go`, `cluster/prov_k8s_prx.go`) found concrete correctness bugs, independently verified against the actual code and vendored `client-go@v0.31.3`. This is a narrow hardening pass — not a redesign — per repo law scope discipline (see `doc/implementation/DEVELOPMENT_LAWS.md`).

Related/broader feature work: #1497 (K8s operator convergence with OpenSVC) — that issue covers StorageClass selection, secrets, dbjobs sidecar, HTTPS bootstrap, GUI manifest view, and proxy-type-aware image selection. This issue is scoped to fixing bugs in the *existing* lifecycle, not adding those capabilities.

Fixed by PR #1738.

## Status

All bugs in the checklist below are fixed and unit-tested on branch `k8s-prov` (PR #1738). Two findings from the original audit — the bootstrap auth header gap and proxy-type-aware image selection — turned out to be out of this issue's scope (they require new capability, not a bug fix) and are tracked under #1497 instead; see "Scope" below. The issue stays open until real Kubernetes integration/regtest coverage exists (T13 debt) — `regtest/test_k8s_provision_scheduler_volume_binding.go` was added and manually verified against a live `kind` cluster, but there is still no CI harness to run it automatically.

## Confirmed bugs (fixed)

- [x] `K8SUnprovisionDatabaseService`: calls `client.AppsV1()` before checking the `K8SConnectAPI()` error — nil-pointer panic on connect failure (F2 violation).
- [x] `K8SGetNodes`: ignores `List()` error and indexes `n.Status.Addresses[0]` unguarded — panics on any node with no reported addresses.
- [x] Port parsing (`strconv.Atoi(s.Port)` / `prx.GetPort()`) silently discards the conversion error in both DB and proxy provisioning — can silently create a Service/container with port 0.
- [x] DB init-container `Command` is a single shell string in a `[]string` instead of `["sh", "-c", "..."]` — malformed for Kubernetes container exec semantics, bootstrap never actually runs correctly.
- [x] `errorChan` contract violations: `K8SUnprovisionDatabaseService` can send 2-3 values on `cluster.errorChan` (per-step error(s) plus an unconditional trailing `nil`) against callers in `cluster/prov.go` that each read exactly once — causes cross-talk with unrelated operations reading the same channel next.
- [x] `K8SUnprovisionProxyService` is a no-op that reports success while leaving the Deployment created by `K8SProvisionProxyService` running — resource leak, false success.
- [x] Proxy Deployment name/selector (`cluster.Name + "-deployment"`, label `app: repication-manager` only) are not unique per proxy — a second proxy in the same cluster collides.
- [x] Namespace/PVC/Deployment/Service create paths lack typed `apierrors.IsAlreadyExists` handling; failed `Create()` results are logged as successes (empty-name objects) instead of stopping/reporting failure.
- [x] `RestartDatabaseService` (generic stop→wait-failed→start since commit 8719ab4f1) silently no-op'd the stop phase for K8S and timed out in `WaitDatabaseFailed` after `MonitorWaitRetry` ticks — restart/rolling-restart/rolling-upgrade were effectively broken for K8S DBs, not just "unsupported". **Fixed via an explicit error, not a scale-based real stop** — `K8SStopDatabaseService` now returns an error immediately instead of silently succeeding, so `RestartDatabaseService`/`StopDatabaseServiceClean`/rolling upgrades fail fast instead of timing out. This is a user-visible behavior change for existing Kubernetes-orchestrated deployments — see PR #1738's description for detail.

## Scope

Hardening only: fix nil-deref/panic risk, explicit error propagation, single-terminal-send on `errorChan`, typed AlreadyExists/NotFound handling, unique proxy naming, explicit port-parse failures, correct init-container command encoding, and implement proxy unprovision for the resource it actually creates.

Out of scope, tracked under #1497 instead (these need new capability, not a bug fix, so they were removed from the checklist above rather than left unresolved here):
- Bootstrap wget sends no auth header, so `api-credentials-secure-config=true` 403s the config fetch. A warning is now logged when this combination is detected at provision time, but closing it requires injecting a credential into the pod (e.g. a Kubernetes Secret).
- `K8SProvisionProxyService` uses `ProvProxProxysqlImg` unconditionally regardless of actual proxy type — native K8s proxy provisioning remains ProxySQL-only until proxy-type-aware image selection exists.

Also explicitly deferred (documented, not fixed here, tracked under #1497): ConfigMap fate (mounted vs dead code), PVC/Namespace retention semantics on unprovision, DB/proxy start/stop as a real K8S lifecycle operation, proxy Service exposure. These require a design decision.

## Validation

Focused Go unit tests using a `kubernetes.Interface` fake-client seam (port validation, naming, AlreadyExists/NotFound classification, nil-safety, single errorChan send, and — via the extracted pure builder `k8sDatabaseDeployment` — the `NodeSelector`-vs-`NodeName` placement fix). `regtest/test_k8s_provision_scheduler_volume_binding.go` reprovisions a server and verifies its PVC binds and pod reaches `Running` against a real Kubernetes API; manually verified passing against a live 3-node `kind` cluster, but has no CI harness to run automatically yet (T13 debt).

caffeinated92:
Closes #1737

Hardens the native Kubernetes provisioning path and adds regression coverage — both fake-client unit tests and a real-cluster regtest — for the scheduler/PVC-binding lifecycle.

## ⚠️ Behavior changes for existing Kubernetes-orchestrated deployments

- **`K8SStopDatabaseService` now returns an explicit error instead of silently succeeding (`nil`).** There is no scale-to-zero/drain semantic implemented, so this was always a no-op — it just used to report success. `RestartDatabaseService`, `StopDatabaseServiceClean`, and the rolling-upgrade paths (master and slaves) all call `StopDatabaseService` with no Kubernetes-specific handling, so **restarts and rolling upgrades against Kubernetes-orchestrated clusters now hard-fail immediately, where they previously appeared to succeed** while doing nothing to the actual pod. This is the correct behavior — reporting success for an action that didn't happen was the actual bug — but it is a visible change for anyone relying on those code paths against a `prov-orchestrator=kube` cluster today.
- **`RestartDatabaseService` for Kubernetes is now a rollout request, not a synchronous readiness wait.** `/actions/restart` patches the Deployment (`restartedAt` + current `ImagePullPolicy`) and returns/clears its restart cookie once that patch succeeds; Kubernetes then replaces the pod asynchronously. This avoids stalling the monitor loop on rollout convergence, but it means success here means "restart requested" rather than "new pod is already Ready".
- Database Deployment pod placement changed from `Spec.NodeName` to `NodeSelector` on `kubernetes.io/hostname`. Functionally equivalent (same node pinning) for clusters using `Immediate`-binding StorageClasses, but **required** for `WaitForFirstConsumer` (the default for most dynamic provisioners, including `kind`'s local-path-provisioner and typical cloud CSI drivers) — previously the PVC would hang in `Pending` forever and the pod would never leave `Init:0/1`. One consequence: pod placement now goes through normal scheduler predicates (taints, etc.), which raw `NodeName` previously bypassed.
- Proxy Deployment name/selector changed from a single shared `<cluster>-deployment` (colliding across all proxies in a cluster) to unique-per-proxy `<cluster>-<proxy>-deployment`. A pre-existing shared deployment from before this change is not touched automatically (ownership can't be proven from a single-proxy-scoped call) — repman now logs a warning if one is detected; see the doc below for the manual migration step.
- `K8SUnprovisionProxyService` now actually deletes the Deployment it's asked to unprovision — previously a no-op that reported success while leaving it running.

All of the above are documented in detail in `doc/implementation/cluster/KUBERNETES_PROVISIONING.md`.

## Bug fixes
- Fixes nil-pointer panic in `K8SUnprovisionDatabaseService` when API connect fails.
- Fixes panic in `K8SGetNodes` when a node has no reported addresses.
- Fixes silent port-0 Services/containers by propagating `strconv.Atoi` errors.
- Fixes DB init-container command encoding (`["sh","-c","..."]`) so bootstrap actually runs.
- Fixes `api-credentials-secure-config=true` bootstrap fetches by sending an `Authorization: Basic` header from the Kubernetes init-container when config auth is enabled; the residual risk (credential is recoverable from the Deployment spec by anyone with RBAC read access) is documented in `doc/implementation/cluster/KUBERNETES_PROVISIONING.md`.
- Enforces single terminal send on `cluster.errorChan` and removes cross-talk from `K8SUnprovisionDatabaseService`.
- Implements `K8SUnprovisionProxyService` and gives proxies unique names/selectors (previously a shared name/selector across every proxy in a cluster).
- Adds typed `apierrors.IsAlreadyExists`/`IsNotFound` handling instead of logging empty-name objects as success.
- Namespace creation is now tolerant of a least-privilege RBAC setup that only grants namespaced-resource permissions, not `namespaces/create`.

Note: proxy image selection (`ProvProxProxysqlImg`) is still unconditional regardless of proxy type — that gap is *not* fixed here, only documented (see `doc/implementation/cluster/KUBERNETES_PROVISIONING.md`); native K8s proxy provisioning still only really supports ProxySQL.

## Refactoring
- Extracts `k8sDatabaseDeployment` as a pure builder (no API calls, no `ServerMonitor` methods invoked) so placement, selectors, and environment — in particular the `NodeSelector` fix above — can be asserted directly in a fast unit test, not only via the live-cluster regtest.

## Regression coverage
- `cluster/prov_k8s_test.go`: fake-client unit tests for nil-safety, port parsing, naming/idempotency, error classification, and — via `k8sDatabaseDeployment` — a direct assertion that the Deployment uses `NodeSelector` (not `NodeName`).
- `regtest/test_k8s_provision_scheduler_volume_binding.go`: reprovisions a server and verifies its PVC binds and its pod reaches `Running` against a real Kubernetes API — this specific check requires a live cluster (fake clientsets don't simulate the scheduler or volume-binding controller) and has no CI harness to run in yet (tracked as T13 debt); manually verified passing against a live 3-node `kind` cluster.

## Deferred
ConfigMap fate, PVC/namespace retention semantics, real K8s start/stop lifecycle, proxy Service exposure, and proxy-type-aware image selection remain design decisions outside this scope, as documented in `doc/implementation/cluster/KUBERNETES_PROVISIONING.md`.

### DIFF
--- a/cluster/cluster.go
+++ b/cluster/cluster.go
@@ -388,6 +388,11 @@ type Cluster struct {
 	opensvcPoolInfoCache   []opensvc.PoolInfo `json:"-"`
 	opensvcPoolInfoCacheAt time.Time          `json:"-"`
 	opensvcPoolInfoCacheMu sync.RWMutex       `json:"-"`
+	// k8sNodeHostnameLabels caches node.Name -> the node's "kubernetes.io/hostname"
+	// label value, captured during K8SGetNodes' nodes/list so Kubernetes
+	// Deployment NodeSelector placement doesn't need a per-node nodes/get call.
+	k8sNodeHostnameLabels   map[string]string `json:"-"`
+	k8sNodeHostnameLabelsMu sync.RWMutex      `json:"-"`
 	// Per-cluster preserved variables (replaces ProvDBConfigPreserveVars mechanism)
 	preservedVars               map[string]string          `json:"-"`
 	preservedVarsExcludeServers map[string]map[string]bool `json:"-"` // varName -> {serverID -> true}
--- a/cluster/cluster_acl_rules.go
+++ b/cluster/cluster_acl_rules.go
@@ -26,7 +26,7 @@ var databaseACLRules = []ACLRule{
 	{"/actions/run-jobs", nil, []string{config.GrantClusterProcess}},
 	{"/actions/provision", nil, []string{config.GrantProvDBProvision}},
 	{"/actions/update-opensvc-template", nil, []string{config.GrantProvDBProvision}},
-	{"/service-opensvc", nil, []string{config.GrantProvDBProvision}},
+	{"/service/", nil, []string{config.GrantProvDBProvision}},
 	{"/actions/unprovision", nil, []string{config.GrantProvDBUnprovision}},
 	{"/actions/start", nil, []string{config.GrantDBStart}},
 	{"/actions/stop", nil, []string{config.GrantDBStop}},
--- a/cluster/cluster_acl_test.go
+++ b/cluster/cluster_acl_test.go
@@ -1245,3 +1245,42 @@ func TestIsURLPassACLOpenSVCPools(t *testing.T) {
 		})
 	}
 }
+
+// The database service/manifest route was generalized from a literal
+// "/service-opensvc" path to a dynamic "/service/{orchestrator}" one
+// (server/api_database.go, #1497 gap 6) -- the databaseACLRules pattern for
+// it (cluster_acl_rules.go) uses strings.Contains, so it had to move from
+// the old literal segment to the "/service/" prefix or every orchestrator
+// would 403 through IsValidClusterACL even with the right grant. Found via
+// live testing against a real JWT-authenticated request, not caught by
+// buildDatabaseServiceConfigResponse's own tests (server/api_database_test.go),
+// since those deliberately bypass this ACL layer.
+func TestIsURLPassACLDatabaseServiceRoute(t *testing.T) {
+	cluster := setupACLTestCluster()
+	cluster.Name = "test"
+
+	cluster.APIUsers["provision_user"] = APIUser{
+		User:   "provision_user",
+		Grants: map[string]bool{config.GrantProvDBProvision: true},
+	}
+
+	tests := []struct {
+		name     string
+		user     string
+		url      string
+		expected bool
+	}{
+		{"opensvc manifest route allowed with provision grant", "provision_user", "/api/clusters/test/servers/db1/service/opensvc", true},
+		{"kubernetes manifest route allowed with provision grant", "provision_user", "/api/clusters/test/servers/db1/service/kube", true},
+		{"denied without provision grant", "user_no_grants", "/api/clusters/test/servers/db1/service/kube", false},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			result := cluster.IsURLPassACL(tt.user, tt.url, false)
+			if result != tt.expected {
+				t.Errorf("User %s on URL %s: Expected %v, got %v", tt.user, tt.url, tt.expected, result)
+			}
+		})
+	}
+}
--- a/cluster/cluster_get.go
+++ b/cluster/cluster_get.go
@@ -319,8 +319,24 @@ func (cluster *Cluster) GetMysqlServerBinaryPath() string {
 	return fmt.Sprintf("%sd", cluster.Conf.BackupMysqlclientPath) // Add d to the end of the path (mysql to mysqld, mariadb to mariadbd)
 }
 
+// GetDomain returns the DNS suffix appended to a server's short s.Name to
+// build its connectable address (server.Domain / server.Host), gated on
+// prov-net-cni for both orchestrators. OpenSVC's CNI-published
+// ".<namespace>.svc.<domain>" resolves to a real, routable address;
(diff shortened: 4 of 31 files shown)


caffeinated92:
@claude review my PR independently

caffeinated92:
@claude review my PR

Issue #1746, opened 2026-09-01
Error-level logs in hot loops flood the Slack hook — rate-limit or log only on state transition


svaroqui:
Follow-up to a client alert storm (#1744 was only half of it).

Every ERROR/WARN log line is forwarded verbatim to the alert Slack hook (cluster_log.go, `LogSlack.Errorf`). When an error sits in a hot path, the channel is flooded: a leftover `prov-orchestrator=opensvc` without its p12 certificate made `OpenSVCConnect()` log "Cannot load OpenSVC cluster certificate" every 2-3 seconds — ~8K Slack messages per day, drowning real alerts.

The state machine already models this exact condition (WARN0099 opens once with the same message), so the repeated raw log adds nothing: per the monitoring laws, alerts must come from state transitions, and logs must not pollute the alert channel.

Proposed fix (pick or combine):
- in `OpenSVCConnect()`: log the cert failure at Err only when WARN0099 transitions to open (`failLoadP12Cert` false→true), Dbg otherwise — the pattern already half-exists via the `failLoadP12Cert` flag;
- generally: rate-limit identical messages to the Slack sink (dedupe key = level+cluster+message, min interval per key), so no future hot-loop error can flood the channel regardless of origin.

svaroqui:
Partial fix landed for v3.1.42: PR #1748 (merged as f1d4dab87) gates the p12 certificate error on the WARN0099 state transition — the concrete hot-loop instance no longer floods the alert channels. The issue stays open (no milestone per triage rules) for the generic design flaw: raw ERROR/WARN log lines are still forwarded verbatim to Slack/Teams; a rate-limiting proposal exists on branch fix/slack-log-flood, pending design discussion (throttle safety-net vs states-only alerting).

PR #1748, opened 2026-09-01
fix(orchestrator): log missing OpenSVC p12 certificate only on state transition

svaroqui:
Fixes the concrete alert-channel flood instance behind #1746.

## Problem
When `prov-orchestrator = "opensvc"` is set with a missing/unreadable p12 certificate, `OpenSVCConnect()` logs the failure at LvlErr on every call, and every raw ERROR log line is forwarded verbatim to the alert channels (Slack/Teams). On a production instance this produced ~8K alert messages/day for a single configuration mistake.

The condition is already a tracked state: `WARN0099` opens via the state machine (which alerts once at the transition) and stays open while the file is missing. Re-logging the raw error adds nothing but noise — the fix logs Err only on the `failLoadP12Cert` edge, Dbg afterwards.

## Validation (dev3, reproduction cluster: opensvc + nonexistent p12 + dedicated test webhook)
- **Before** (develop): 17× "Cannot load OpenSVC cluster certificate" forwarded to the alert channel in 6 minutes.
- **After** (this branch): **1** occurrence, then silence; `WARN0099` verified open in `/topology/alerts` with the full description — the state carries the persistent condition.

Also scrubs a client name from a test comment (public-repo hygiene).

Part of #1746 — the generic design flaw (every raw ERROR/WARN log line forwarded to the alert channels) stays open; a separate rate-limiting proposal exists on branch fix/slack-log-flood for discussion.

### DIFF
--- a/cluster/cluster_bck_test.go
+++ b/cluster/cluster_bck_test.go
@@ -2585,7 +2585,7 @@ func TestResticInitRepoWithOptions_ProceedsWhenBucketPreCheckFails(t *testing.T)
 func TestPstates30_IncludesRejoinCatalogStates(t *testing.T) {
 	// WARN0190/0191: HasCatalogBackupForRejoin. WARN0161/0162:
 	// CheckClusterServiceAgents. WARN0169: CheckOnPremiseSSHKey. WARN0170:
-	// CheckConfiguratorPrerequisites (the 2026-09 mixr Slack alert storm).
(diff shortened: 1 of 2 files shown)


svaroqui:
Merged into develop as f1d4dab87 (merge commit landed despite a GitHub 502 mid-transaction; the PR object never flipped to merged). Closing manually — all commits are in develop.

Issue #1749, opened 2026-09-01
Instances with small prov-db-memory can be OOM-killed by the generated configuration


svaroqui:
## Problem

An instance declared with a small `prov-db-memory` can be OOM-killed by the very configuration the configurator generates — either at boot (mariadbd killed before InnoDB init) or at runtime (nightly job windows). The operator declares one number; the generated profile must fit in it, and today it does not.

Observed live: a 768MB-declared dev instance whose freshly rematerialized configuration OOMs mariadbd at every boot (kernel memcg oom-kill, anon-rss ≈ cap), which also broke a rolling restart (both slaves down at boot → no switchover candidate → rolling canceled). Production-side, an 8GB instance was OOM-killed at night with ~3.6GB of RSS above its tracked buffers.

## What contributes (each detailed/measured in comments)

1. **PFS digest sizing** — the `pfs` tag template hardcodes `performance_schema_max_{digest,digest_text,sql_text}_length = 16384` (defaults 1024): P_S preallocates **727MB measured** at init regardless of prov-db-memory. The proven boot-time killer.
2. **Redo log size** — `GetConfigInnoDBLogFileSize()` floors at 1024MB: the generated redo can exceed the whole cgroup. (The `smallredolog` tag also has two divergent implementations: Go getter 128M vs moduleset template 64M, settled by include-file alphabetical order.)
3. **Session memory is uncapped** — `max_session_mem_used` is left at its unlimited default while `max_connections=1000`, and MariaDB's own pressure-driven reclamation is disabled in our containers (every boot logs `memory.pressure not writable, functionality unavailable to MariaDB`).
4. **No allocator headroom** — the memory split budgets 100% to caches; glibc arena fragmentation alone accounted for the production gap. jemalloc-by-default is in flight (PR #1751); the split still needs a reserve (~20%).

## Direction (maintainer-approved)

Derive every consumer from `prov-db-memory` — no absolute memory constants in templates: PFS lengths scaled (or a dedicated `pfscapture` tag), redo = min(BP/2, mem/4) floor 64M, generate `max_session_mem_used`, make `memory.pressure` writable, keep a headroom reserve.

## Acceptance test

An instance declared at 768MB must boot healthy on purely generated configuration, no overrides. A live reproduction is kept on a dev cluster.

svaroqui:
[...]

PR #1751, opened 2026-09-01
feat(prov): jemalloc preload + derived MALLOC_ARENA_MAX in the db container service sheet

svaroqui:
Part of #1749 (the OOM incidents: glibc arena fragmentation inside tight cgroups).

## What

The OpenSVC `container#db` environment now exports allocator tuning for MariaDB/MySQL containers:

- `LD_PRELOAD=<prov-db-docker-jemalloc-preload>` — new setting, default `libjemalloc.so.2` (shipped by the MariaDB images — **verified live on mariadb:11.4**: library present, preload maps jemalloc, no loader error). The soname/path is configuration so the library version follows the image, never the code. Empty value disables both exports (off-switch).
- `MALLOC_ARENA_MAX=<prov-cores>` — **derived, not a new knob**: arena count is a lock-contention parameter that scales with the real parallelism the cgroup can run (the same prov-cores that produces `--cpus`), not with memory or connections. Fallback to 2 when prov-cores is unparseable.

Fallback chain: image ships jemalloc → jemalloc handles the heap (returns freed pages to the kernel, immune to the arena ratchet); image lacks it → ld.so logs one warning, glibc stays and honors the derived MALLOC_ARENA_MAX.

## Scope

OpenSVC docker/podman provisioning (V2/V3 service sheet) AND the kubernetes DB deployment — both consume the shared `GetDBAllocatorEnv` (cluster/prov.go), so the two orchestrators provision the same default. DB container only — jobs/init/sidecar containers run scripts, not the server. Onpremise provisioning is a separate path, not touched. Note: the active k8s-prov branch also extends prov_k8s_db.go — small conflict possible, coordinate merge order.

## Validation

- Unit test `TestOpenSVCGetDBContainerEnvironment` (derivation, verbatim preload path, fallback, off-switch).
- Image check on a dev host: `mariadb:11.4` has `/usr/lib/x86_64-linux-gnu/libjemalloc.so.2`; `LD_PRELOAD` by soname maps it (5 mappings, no stderr).
- Full effect on a live cluster requires a reprovision/restart of the db service — planned on a dev cluster before this lands anywhere real.

Settings are reachable through the standard cluster settings API/GUI mechanism (string + implicit off via empty value).

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -517,4 +517,4 @@ When creating new implementation documentation:
 
 **Docker Rootless Permissions**: When running rootless Docker containers, ensure mounted volumes have correct ownership (`chown 1000:1000` or appropriate UID/GID for the `repman` user).
 
-**Environment Variable Conflicts**: If using both TOML config and environment variables, remember that environment variables take precedence over TOML values. Use `replication-manager config-merge` to debug effective configuration.
+**Environment Variable Conflicts**: If using both TOML config and environment variables, remember that environment variables take precedence over TOML values. To inspect the effective configuration, read it from the API (`/api/clusters/{name}` config object). `config-merge` is NOT an inspection tool: it consolidates dynamic-mode changes back into the original config file and OVERWRITES it.
--- a/cluster/prov.go
+++ b/cluster/prov.go
@@ -10,6 +10,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"strconv"
 	"strings"
 	"sync"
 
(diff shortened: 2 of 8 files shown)


svaroqui:
**Live validation on a dev cluster (OpenSVC, mariadb:10.11):**

- New setting visible in the runtime cluster config after deploying the branch: `provDbDockerJemallocPreload: libjemalloc.so.2` (default applied, no config change needed).
- Applied the exact generated environment line to the db service sheet (`container#db.environment = MYSQL_INITDB_SKIP_TZINFO=yes LD_PRELOAD=libjemalloc.so.2 MALLOC_ARENA_MAX=2`, arena=2 from the prov-cores fallback) and restarted the db container via `om restart --rid container#db` — instance back up in 2.2s.
- Runtime proof on the mariadbd process: both variables in `/proc/<pid>/environ`, **6 jemalloc mappings in /proc/<pid>/maps** — jemalloc is the active allocator; topology healthy after restart (Master + 2 Slaves).
- Fallback path checked on `mysql:8.0.41-debian` (no jemalloc shipped): the loader prints a single 'ignored' warning and the process runs on glibc, where MALLOC_ARENA_MAX applies.

Caveat: the end-to-end `update-opensvc-template` API push could not be exercised on this instance — its OpenSVC v3 API auth is broken independently of this PR (ERR00082 'no access or refresh tokens', pre-existing: the p12 secret doesn't decrypt with the local key). The template generation itself is unit-covered; the sheet application was done with the same content via `om set`.

Issue #1760, opened 2026-09-02
DBU as a first-class backend concept: ChangeDBUnit orchestrator + client resize script


svaroqui:
## Context
The DBU (Database Unit) model lives only in the **frontend** today: the ratios `1 DBU = 1 core / 4096 MB / 40 GB / 1000 IO/s` are hardcoded in the DBUSlider JSX (`share/dashboard_react/src/Pages/Configs/components/DBConfigs.jsx`), and changing the DBU dispatches **4 raw `setSetting` calls**. There is no backend DBU concept.

Consequences observed:
- The generic `setClusterSetting` (server/api_cluster.go) only **WARNs then overwrites** an immutable (pinned) `prov-*` spec, so a DBU change silently overrides a pinned `prov-db-memory` — contradicts "pinned = hard lock".
- No **client-overridable** hook to apply a live resource change (dynamic cgroup); today a memory change posts a reprov cookie -> full container recreation.
- Business logic (DBU ratios, billing floor) sits in the GUI.

Already handled on #1749 (branch `fix/prov-db-memory-budget`): the Configurator display cap (gauge no longer clamps the shown value) and the plan-attach path (`SetServicePlanInfos` now respects immutable specs). **This issue covers the backend DBU refactor.**

## Design
Single backend entry point:
```go
func (cluster *Cluster) ChangeDBUnit(dbu int) error
```
- **`CanIncreasePlan(dbu)` / `CanDecreasePlan(dbu)`** — feasibility gates (host capacity, quota, peer health, billing floor; safe-to-shrink = current RSS < new limit).
- **`InPlaceMigrateDBUnit(from, to)`** — applies specs **respecting pins** (immutable = untouched), preserves the DBU billing floor, and **calls `prov-db-resize-script` if non-empty** (client-overridable live resize: `docker update --memory` / `om set`), else falls back to the reprov path. **Directional ordering**: grow = cgroup up then `SET GLOBAL innodb_buffer_pool_size` up; shrink = BP down then cgroup down (avoid OOM).
- DBU->resource ratios move to the backend (constants, later configurable in TOML per T16).
- New API `POST /api/clusters/{name}/actions/set-dbu/{n}` -> `ChangeDBUnit`; the GUI DBUSlider calls it instead of 4 `setSetting`.
- New hook **`prov-db-resize-script`** following the `cluster/prov_scripts.go` moule; **off by default** (empty = disabled, T14).

## Billing-floor semantics
[...]

PR #1761, opened 2026-09-02
serviceplan: refuse plan change when a plan-driven spec is pinned (#1760)

svaroqui:
First slice of #1760.

A service-plan (DBU) change drives `prov-db-{memory,cpu-cores,disk-size,disk-iops}` and `prov-proxy-{cpu-cores,disk-size}` plus the DB variables they map to. If the operator has pinned any of them, the plan can no longer move that spec — so the whole plan change is refused rather than silently partially applied.

- **`CanChangePlan()`** returns false when any plan-driven prov flag is **immutable** (immutable.toml) OR any driven DB variable is **preserved** (01_preserved.cnf, the "agree"). New `IsVariablePreserved()` membership helper.
- **`SetServicePlan`** refuses up front with a clear error; the `prov-service-plan` API now **returns** that error (it was discarded) so the refusal surfaces to the GUI.

Built-in default gate. The **client-overridable `prov-db-change-plan-script` hook** (bash, prov_scripts.go moule, exit≠0 = refuse) is the planned override, tracked in #1760 along with the rest of the DBU backend refactor.

Note: on a cluster with a pinned `prov-db-memory` (e.g. dev3 at 768M), plan changes are now correctly refused until it is unpinned.


svaroqui:
Closing: decided not to apply this PR.

Issue #1762, opened 2026-09-02
Missing 'force-agree' flag: a regenerated config isn't reconciled to the running DB without a full reprovision (02_delta override)


svaroqui:
## Problem
When the generated config changes (a `prov-db-memory` change, a moduleset/compliance update, a binary upgrade), repman regenerates `replication-manager.d/*.cnf` with the NEW values, but the node keeps **running the OLD values**. The difference is captured by `WriteDeltaVariables` into `custom.d/02_delta.cnf` (`# delta:value-differs`), and since `custom.d` is loaded **after** `replication-manager.d`, the old values **override** the new config at runtime.

`prov-auto-update-compliance` accepts the new module/template and regenerates the config (`AcceptComplianceUpdate`, cluster_version.go), but it operates at the **module/template** level — it does **not** reconcile the per-variable delta (`02_delta`). For a **known** variable there is no auto-agree (only **unknown** variables auto-route to `03_agreed` for review, srv_cnf.go). So the node stays on the old profile.

## The three config layers (custom.d)
- `01_preserved.cnf` — operator explicitly preserves (keep running).
- `02_delta.cnf` — auto list of "differs, but neither preserved nor agreed" (undecided drift). **Loaded → overrides the generated config, keeping the old values.**
- `03_agreed.cnf` — accepted values pushed to the DB (the apply channel).

## Why a restart doesn't fix it
A plain restart re-reads `02_delta.cnf` and re-applies the old values → the delta **self-perpetuates** (running stays old → running≠config → delta rewritten old). Verified: after a restart, `performance_schema_max_digest_length` stays at the old value and `02_delta.cnf` is still full.

## Current reconciliation paths (both heavy/manual)
- `WipeDeltaConfig()` is only called from **`InitDatabaseService`** (reprovisioning, prov.go) → a full reprovision, which can reseed a slave.
- No API/flag to "accept the pending delta and apply the generated config" without a reprovision.

## Missing capability
A flag to **force-agree the deltas** — the variable-level counterpart of `prov-db-force-write-config` (which forces the config **files** to be replaced instead of written as `.new`). When set, it would agree the pending `02_delta` variables (move to `03_agreed` / wipe the delta), so the next restart boots on the new generated config — **without** a full reprovision. Must respect `01_preserved` (never override an operator-preserved variable).

## Repro
1. Change a generated value (e.g. lower `prov-db-memory` so the configurator emits a smaller buffer pool / a lower `performance_schema_max_digest_length`).
[...]

Issue #1495, opened 2026-05-29
feat: consolidated single-file config from tags with version-aware variable filtering


svaroqui:
## Summary

Replace the current multi-fragment `conf.d/` symlink approach with a single consolidated config file generated by the configurator. Tags still define fragments, but the output is a single `replication-manager.cnf` containing only variables recognized by the target database version.

## Motivation

- `loose_` prefix hides deprecated and unknown variables silently — no visibility into what's actually applied
- Multiple symlinked fragments make debugging config issues harder
- File ordering with numbered prefixes (`10_engine.cnf`, `20_replication.cnf`) is fragile
- No way to know if a compliance tag variable is ignored by the DB without manual checking

## Proposed Architecture

```
Tag fragments (source)          Consolidated output (deployed)
─────────────────────           ──────────────────────────────
replication-manager.d/          conf.d/
├── 10_engine.cnf               └── replication-manager.cnf  ← single symlink
├── 20_replication.cnf              (merged, filtered, clean)
├── 30_security.cnf
├── 40_pfs.cnf
└── ...
```

### Build steps

1. **Merge** all tag fragments in order (respecting numbered prefixes for last-wins)
2. **Query DB binary** — `mariadbd --help --verbose` to get the list of recognized variables for this exact version
3. **Filter** — remove any variable not in the recognized list (no more `loose_` needed)
4. **Report** — flag removed variables in the configurator UI (unknown/deprecated badge)
5. **Deploy** — single `replication-manager.cnf` symlinked into `conf.d/`

### Benefits

- Clean config — no `loose_`, no deprecated clutter, no silent ignores
- Version-aware — the DB binary is the source of truth for what's valid
- Single file — easier debugging, no fragment ordering issues
- Full visibility — every filtered variable is surfaced in the configurator
- Foundation for goal 5 (bug-aware config) — same filtering mechanism can disable buggy features per version range

### Backward compatibility

- Fragments in `replication-manager.d/` remain for tracking and tag visualization
- `custom.d/` files (`01_preserved.cnf`, `02_delta.cnf`) still override the consolidated file
- Existing clusters can migrate gradually — flag to switch between fragment mode and consolidated mode

### Depends on

- dbjobs unknown variable detection (PR #1494)
- Config tracking agreed/delta/preserved flow (PR #1494)

## Milestone

Target: v3.2

PR #1763, opened 2026-09-03
Agree Compliance: drop agreed on reconcile + opt-in auto-agree (#1762)

svaroqui:
Implements the two parts of #1762 (the DB-side counterpart of Auto-Update Compliance).

## Part 1 — drop agreed on reconcile (lifecycle fix)
An agreed variable (`03_agreed.cnf`) only exists to hold a value until the DB restarts with the compliance value. Once the DB actually runs it (`Runtime == Config`), nothing cleared the agree, so `HasConfigDiff` stayed true and the GUI showed a **phantom diff** on a node fully reconciled at the DB level.

- `VariablesMap.DropReconciledAgreed()`: drops the agreed marking of every variable running the compliance value; rewrites `03_agreed.cnf`. **Never** touches operator-forced preserves (`PreservedSource != ""`, i.e. `01_preserved.cnf` / cluster-level) — those are deliberate overrides that must survive.
- Called in the monitor cycle right after the runtime refresh, before `HasConfigDiff`.

## Part 2 — prov-auto-agree-compliance (opt-in auto-agree)
Update-Compliance is repman-side only (regenerates config); it never reaches the DB. This adds the DB-side auto-agree.

- New per-cluster flag `prov-auto-agree-compliance` (**off by default**, T14).
- `VariablesMap.AutoAgreeValueDeltas()`: for each **value-differs** delta (known variable, `Config != nil`, `Deployed != Config`, not preserved, not dropped) sets `Preserved = Config` → routes to `03_agreed.cnf` → the DB adopts the compliance value on its next restart; `DropReconciledAgreed` then clears it. **Scoped to value-changes only** — deprecated/unknown drops are left for manual review because `loose_` hides them (blocked on #1495).
- Wired in the monitor cycle on deployed-config change, guarded by the flag.

## Key design point
Agreed vs forced are distinguished: agreed = `Preserved != nil && PreservedSource == ""` (`Preserved == Config`); forced = `PreservedSource != ""` (`Preserved != Config`, `IsPreserved()`). The clean-on-reconcile only drops agreed, never forced.

Tests: `TestDropReconciledAgreed`, `TestAutoAgreeValueDeltas`.

Remaining (follow-up): GUI toggle for the flag (T6) + user/implementation docs (T8).

### DIFF
--- a/cluster/srv.go
+++ b/cluster/srv.go
@@ -1174,6 +1174,11 @@ func (server *ServerMonitor) Refresh() error {
 				// Clear the flag after processing
 				server.VariablesMap.ClearDeployedChanged()
 			}
+		} else {
+			// Config tracking off (prov-db-config): clear the drift pill so the GUI
+			// (hasConfigDiff) reflects the live off-state — surfaced as CINF0003 —
+			// instead of a value frozen from when tracking was last on.
+			server.HasConfigDiff = false
 		}
 
 		if server.IsNeedPathCheck {
--- a/cluster/srv.go
+++ b/cluster/srv.go
@@ -1148,12 +1148,12 @@ func (server *ServerMonitor) Refresh() error {
 				}
 
 				// Auto-agree value-differs deltas to the compliance value (opt-in,
-				// prov-auto-agree-compliance — the DB-side counterpart of
+				// prov-db-compliance-auto-agree — the DB-side counterpart of
 				// prov-auto-update-compliance). Runs before the delta is (re)written so
 				// agreed variables route to 03_agreed instead of 02_delta and the DB
 				// adopts them on its next restart. Value-changes only; deprecated /
 				// unknown drops are left for manual review (loose_ hides them, #1495).
-				if cluster.Conf.ProvAutoAgreeCompliance {
+				if cluster.Conf.ProvDBComplianceAutoAgree {
 					if agreed := server.VariablesMap.AutoAgreeValueDeltas(); len(agreed) > 0 {
 						cluster.LogModulePrintf(cluster.Conf.Verbose, config.ConstLogModGeneral, config.LvlInfo,
 							"Auto-agreed %d value-differs delta(s) to compliance on %s: %s", len(agreed), server.URL, strings.Join(agreed, ", "))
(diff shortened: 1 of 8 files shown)


svaroqui:
@claude please review this PR.

Summary of what it does (fixes #1762):
- `prov-db-compliance-auto-agree` (opt-in, off by default): auto-agree value-differs config deltas to the compliance value; value-changes only — dropped/unknown variables are never auto-agreed (they'd crash the DB on restart, and `loose_` makes them undetectable — see #1495).
- `DropReconciledAgreed`: once the DB runs the compliance value (Runtime == Config), the reconciled agreed entry is dropped so it stops showing as a phantom diff. Operator-forced preserves are never touched. Gated on a runtime change so the steady state does no extra map walk (F2).
- Gate the whole per-tick config-diff / compliance block behind the existing `prov-db-config` master switch (T14 client off-switch, F2 no monitor burden).
- GUI: Auto-Agree Compliance toggle in the Configurator tab + backend switch case.

Please look especially at:
1. The reconcile/agree scoping in `config/maps.go` (DropReconciledAgreed / AutoAgreeValueDeltas) — correctness of the Preserved/PreservedSource routing and the UNKNOWN safety (agreed:unknown-variable never auto-agreed/dropped).
2. The monitor-cycle wiring and gating in `cluster/srv.go` (brace placement, HasConfigDiff kept live-per-tick semantics, runtimeChanged gate).
3. Any missed off-switch or state-tracking law violation.

svaroqui:
@claude the previous review hit the 40-turn limit. Please do a **tightly scoped** review this time — look **only** at the core logic, and skip the docs, the GUI (jsx), and the flag rename entirely.

Review just these two files:

1. `config/maps.go` — the two new methods:
   - `DropReconciledAgreed()`: drops an agreed entry (Preserved != nil && PreservedSource == "") once Runtime == Config. Verify it never touches operator-forced preserves (PreservedSource != "") nor unknown-variable entries (PreservedSource == "unknown-variable").
   - `AutoAgreeValueDeltas()`: agrees value-differs deltas (Preserved == nil && !Dropped && Config != nil && Deployed != Config) to Config. Verify UNKNOWN (Config==nil / already-Preserved) and Dropped are never auto-agreed.
   - `SetRuntimeValues()` now returns the changed-count.

2. `cluster/srv.go` (monitor cycle, ~L1104–1181) — the wiring: the whole VariablesMap block gated behind `prov-db-config`; the drop gated on runtimeChanged>0; HasConfigDiff / HasDeployedChanged unaffected; brace placement correct.

Focus on correctness and any missed edge case. One concise pass, then post — don't explore the rest of the PR.

PR #1770, opened 2026-09-05
fix(prov): serialize errorChan provisioning ops with a per-cluster mutex (#1769 phase 1)

### LINKED ISSUE #1769: bug(prov): shared cluster.errorChan cross-talk deadlocks rolling ops, leaves server stuck in maintenance
## Summary

`cluster.errorChan` is a **single, unbuffered, cluster-scoped** channel (`cluster/cluster.go:314` decl, `:598` `make`) shared by **every** provisioning IPC path: 9 `<-cluster.errorChan` receivers in `cluster/prov.go`, plus `ServerMonitor.Uprovision()` (`cluster/srv.go:1851`), against senders spread across `prov_opensvc_db.go`, `prov_k8s_prx.go`, `prov_localhost_*.go`.

There is **no correlation** between a given operation and the result it reads, and **no mutex** serializing provisioning ops (`prov.go:1113 cluster.Lock()` is the general cluster lock, not a provisioning gate). When two provision / unprovision / config-send operations **overlap in time**, their sends and receives cross: one receiver picks up the *other* op's result, and the **other receiver blocks forever**. That deadlock halts orchestration and leaves the affected server stuck — e.g. in **maintenance** (set before the blocking call, cleared only after it returns) — **with no log**, because the blocked call emits nothing while it waits on the channel.

## Relation to #1737 (not a duplicate — this is the root)

#1737 already identified this exact failure mode ("cross-talk with unrelated operations reading the same channel next") and fixed the **K8s unprovision sender** to send exactly once. That hardening removes multi-send cross-talk *within a single k8s op*. It does **not** address the root: even with every sender sending exactly once, **two concurrent ops still cross-talk** on the one shared channel. OpenSVC and localhost paths, and the fundamental design, remain fully exposed.

## Root cause

- One shared unbuffered `cluster.errorChan` for all provisioning IPC (no per-op channel).
- No op↔result correlation; no serialization of provisioning ops.
- => any temporal overlap of two `errorChan` users = one of them deadlocks.

## Repro (OpenSVC, observed on a dev cluster)

A rolling reprovision (`RollingReprov` → `UnprovisionDatabaseService`, `prov.go:451` reads `<-cluster.errorChan`) ran while a **config resend** on the same server (`OpenSVCUpdateDatabaseServiceConfig`, which sends on the same channel at `prov_opensvc_db.go:486` and `:643`) was triggered by a config change. The config-send consumed the unprovision goroutine's single send; the reprov's `<-cluster.errorChan` never received → the reprov blocked indefinitely → the server was left **in maintenance** with no further log. Timeline confirmed by the config-send log lines landing exactly when the server went silent.

## Blast radius

Infra-wide across the whole rolling family — **restart, upgrade, and reprov** all funnel through provisioning ops that read this channel, and a monitor-triggered config resend can overlap any of them. **Not reprov-specific.** OpenSVC and localhost paths fully exposed; k8s partially mitigated by #1737's single-send but still exposed to concurrent-op cross-talk.

## Severity

**Blocker** — a deadlock that halts orchestration/monitoring (functional law 2: never compromise monitoring).

## Proposed fix (design — for team discussion before coding)

Replace the shared `cluster.errorChan` with a **per-operation channel**: the caller allocates a local `chan error` (buffered, cap 1), passes it into the goroutine, and reads only its own result. Removes cross-talk by construction, preserves concurrency, and **supersedes** the per-path single-send hardening. Alternative — serialize all provisioning ops under one mutex — is simpler but kills concurrency; not preferred. Coordinate with the active k8s work (#1737 / k8s-prov) before implementing.

svaroqui:
## What

Phase 1 mitigation for #1769: serialise the shared, unbuffered
`cluster.errorChan` provisioning IPC with a per-cluster mutex.

## Why

`cluster.errorChan` (`make(chan error)`, unbuffered, cluster-scoped) is used by
~10 provision/unprovision receivers (`cluster/prov.go` + `cluster/srv.go`
`Uprovision`) with **no op<->result correlation** and **no serialization**.
When two operations on the same cluster overlap, their sends/receives cross:
one receiver reads the other op's result and the other **blocks forever**,
deadlocking orchestration and leaving a server stuck in maintenance with **no
log** (a blocked `<-errorChan` prints nothing). Halts monitoring → release
blocker (F2).

Observed: a rolling reprovision's `UnprovisionDatabaseService` overlapped a
config-build send and hung, leaving the old master in maintenance.

## What this does

Add `cluster.provisioningMutex sync.Mutex`, taken around the launch-and-receive
span of each of the 8 receiver wrappers (`ProvisionServices`,
`InitDatabaseService`, `InitProxyService`, `InitAppService`, `Unprovision`,
`UnprovisionProxyService`, `UnprovisionDatabaseService`,
`ServerMonitor.Uprovision`).

- **Covers** all cross-op cross-talk. The config *update* path
  (`OpenSVCUpdateDatabaseServiceConfig`) does not use `errorChan`, so wrapping
  the receivers covers every channel user.
- **Preserves** fan-out parallelism (per-server goroutines still run in
  parallel under one hold), cross-cluster parallelism, and monitor liveness
  (the monitor's config path does not take this lock).
- **Residual** intra-op multi-send (config-building helpers such as
  `OpenSVCGetDBEnvSection` sending mid-flight) is left to **Phase 2** — the
  per-operation-channel refactor (see the epic on #1769; touches the K8s
  sender files, to be coordinated with the active K8s work).

Design writeup: `doc/implementation/cluster/ERRORCHAN_PROVISIONING.md`.

## Test

- `go build ./cluster/...` green; `go vet` clean (no copylocks — `Cluster` is
  always used by pointer).
- Live validation: rolling reprovision on a dev OpenSVC cluster (T13) — see PR
  thread.

Closes #1769 (Phase 1; Phase 2 tracked as the epic on the same issue).

### DIFF
--- a/cluster/cluster.go
+++ b/cluster/cluster.go
@@ -288,30 +288,43 @@ type Cluster struct {
 	// MaintenanceLogrus is a dedicated logrus.Logger that writes to maintenance.log.
 	// Receives ConstLogModMaintenance events: backup, SST, task execution, purge, etc.
 	// Set by the server on cluster init. Nil when no log-file is configured.
-	MaintenanceLogrus                   *log.Logger                 `json:"-"`
-	runOnceAfterTopology                bool                        `json:"-"`
-	logPtr                              *os.File                    `json:"-"`
-	termlength                          int                         `json:"-"`
-	runUUID                             string                      `json:"-"`
-	cfgGroupDisplay                     string                      `json:"-"`
-	RepMgrVersion                       string                      `json:"-"`
-	RepMgrFullVersion                   string                      `json:"-"`
-	RepMgrRestartTime                   int64                       `json:"-"`
-	RepMgrHostname                      string                      `json:"-"`
-	exitMsg                             string                      `json:"-"`
-	exit                                atomic.Bool                 `json:"-"`
-	stopOnce                            sync.Once                   `json:"-"`
-	canFlashBack                        bool                        `json:"-"`
-	canResticFetchRepo                  bool                        `json:"-"`
-	failoverCond                        *nbc.NonBlockingChan        `json:"-"`
-	switchoverCond                      *nbc.NonBlockingChan        `json:"-"`
-	rejoinCond                          *nbc.NonBlockingChan        `json:"-"`
-	bootstrapCond                       *nbc.NonBlockingChan        `json:"-"`
-	altertableCond                      *nbc.NonBlockingChan        `json:"-"`
-	addtableCond                        *nbc.NonBlockingChan        `json:"-"`
-	statecloseChan                      chan state.State            `json:"-"`
-	switchoverChan                      chan bool                   `json:"-"`
-	errorChan                           chan error                  `json:"-"`
+	MaintenanceLogrus    *log.Logger          `json:"-"`
+	runOnceAfterTopology bool                 `json:"-"`
+	logPtr               *os.File             `json:"-"`
+	termlength           int                  `json:"-"`
+	runUUID              string               `json:"-"`
+	cfgGroupDisplay      string               `json:"-"`
+	RepMgrVersion        string               `json:"-"`
+	RepMgrFullVersion    string               `json:"-"`
+	RepMgrRestartTime    int64                `json:"-"`
+	RepMgrHostname       string               `json:"-"`
+	exitMsg              string               `json:"-"`
+	exit                 atomic.Bool          `json:"-"`
+	stopOnce             sync.Once            `json:"-"`
+	canFlashBack         bool                 `json:"-"`
+	canResticFetchRepo   bool                 `json:"-"`
+	failoverCond         *nbc.NonBlockingChan `json:"-"`
+	switchoverCond       *nbc.NonBlockingChan `json:"-"`
+	rejoinCond           *nbc.NonBlockingChan `json:"-"`
+	bootstrapCond        *nbc.NonBlockingChan `json:"-"`
+	altertableCond       *nbc.NonBlockingChan `json:"-"`
+	addtableCond         *nbc.NonBlockingChan `json:"-"`
+	statecloseChan       chan state.State     `json:"-"`
+	switchoverChan       chan bool            `json:"-"`
+	errorChan            chan error           `json:"-"`
+	// provisioningMutex serialises every provision/unprovision operation that
+	// reports its result through the shared, unbuffered errorChan (the ~10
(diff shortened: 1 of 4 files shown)


svaroqui:
## Live validation (dev OpenSVC cluster, T13)

Deployed this branch to a dev OpenSVC cluster (`v3.1.41-73-g931944068`) and
exercised the exact path that hung in the incident.

| Step | Result |
|---|---|
| repman restart under the new `provisioningMutex` | started clean, no boot deadlock |
| `GET .../servers/db1/actions/unprovision` (calls `UnprovisionDatabaseService` — the path that hung) | executed; db1 service torn down; **repman stayed responsive** |
| `GET .../servers/db1/actions/provision` (takes the **same** mutex) | **HTTP 200 in 16.6s** — proves the unprovision released the lock and lock-holding ops serialize and complete |
| `/api/version` throughout | **~0.0003s** every poll — monitoring never blocked (F2) |

The full unprovision → provision round-trip through the mutex completed with no
deadlock and repman never wedged — the opposite of the reported incident, where
the same unprovision left the server stuck in maintenance and silent.

Note: this confirms the mutex serialises correctly without self-deadlock and
keeps the monitor live. It does not attempt to force the original timing race
(the trigger — a concurrent config-build send — is non-deterministic); that
guarantee is structural (only one op touches the shared channel at a time) and
covered by review + the design writeup.

svaroqui:
## Added: guarantee the reseed on rolling reprovision (#1771)

Commit `5860d6ac4` (on this branch) also fixes **#1771**: `RollingReprov`
destroys each replica's data but delegated the reseed to the autoseed rejoin
path, which is off by default → silent data loss.

Now `RollingReprov`:
- refuses to run if `autorejoin` is disabled (aborts before destroying anything);
- pilots `autoseed` on for the duration (restored afterwards) so each
  reprovisioned replica is reseeded from the master.

Note: this PR now carries two related provisioning fixes (#1769 errorChan mutex
+ #1771 reprov reseed). Say the word and I'll split #1771 into its own PR.

svaroqui:
Update: the reprovision reseed fix (#1771) has been **split out into its own PR (#1772)**. This branch is back to the single errorChan mutex fix (#1769). Sorry for the noise.

svaroqui:
Superseded by #1773 — consolidated into a single rolling-improvement branch (fix/rolling-provisioning-hardening).

PR #1772, opened 2026-09-05
fix(prov): guarantee the reseed on rolling reprovision (#1771)

### LINKED ISSUE #1771: bug(prov): RollingReprov destroys replicas without guaranteeing the reseed (silent data loss when autoseed is off)
## Summary

`RollingReprov` is a **destructive** operation — for each replica it does
`UnprovisionDatabaseService` (destroy the service + volume) then
`InitDatabaseService` (reprovision an **empty** service) then
`StartDatabaseWaitRejoin`. The data restore is **not performed by the reprov
itself**: it is delegated to the monitor's rejoin path, which reseeds a
freshly-provisioned (standalone, no-crash) node **only if `autoseed` is on**.

`RollingReprov` (`cluster/cluster_roll.go`) does **not** pilot, force, or
require `autoseed` (nor `autorejoin`) — it silently depends on the standing
cluster config. `autoseed` defaults to **false** (`server/server.go`:
`flags.BoolVar(&conf.Autoseed, "autoseed", false, "Automatic join a standalone node")`).

## The gate

`cluster/srv_rejoin.go` (standalone node, no crash record — the freshly
reprovisioned case):

```go
crash := cluster.getCrashFromJoiner(server.URL)
if crash == nil {
    if cluster.oldMaster != nil && cluster.oldMaster.URL == server.URL {
        server.RejoinMasterSST(); return
    }
    if cluster.Conf.Autoseed {           // <-- gate
        server.ReseedMasterSST()         // full physical reseed from master
        return
    }
    ...
    // "No crash, no old-master anchor, no autoseed: attach read-only ... on current GTID"
    server.attachAsReadOnlySlave(cluster.master)   // attaches WITHOUT a data reseed
}
```

## Impact

Run `RollingReprov` on a cluster with `autoseed = false` (the default) and the
replicas are **destroyed and reprovisioned empty**, then merely attached
read-only on the current GTID — no historical data. Result: empty/incoherent
replicas (or a stuck `StandAlone`, as observed on a dev cluster where a
reprovisioned node stayed `StandAlone` until reseeded by hand). A destructive
operation whose restore step is not guaranteed by the operation = **silent
data loss on the replicas** when the operator has not separately enabled
`autoseed`.

## Proposed fix

`RollingReprov` must **guarantee** the reseed rather than depend on standing
config. Options (to discuss):

1. **Refuse to run** if the reseed is not configured: pre-flight check that
   `autoseed` (and, where relevant, a reseed method) is enabled; otherwise
   abort with a clear error before destroying anything.
2. **Pilot it for the duration**: enable `autoseed` (save/restore the prior
   value) around the reprov so each reprovisioned replica is reseeded, then
   restore the operator's original setting.
3. **Reseed explicitly**: after `InitDatabaseService`, drive the reseed
   directly instead of relying on the passive rejoin path.

Preference: (1) as a minimum safety gate now (never destroy without a
guaranteed restore), optionally (2) for convenience. This also keeps the user
doc's "rolling reprovision reseeds the replica from the primary" accurate.

Related: the reprov also shares the errorChan deadlock class (#1769).

svaroqui:
## What

Fix #1771: `RollingReprov` destroys each replica's data (`UnprovisionDatabaseService`)
then recreates an empty service, but the reseed that repopulates it was delegated to
the autoseed rejoin path (`RejoinMaster` -> `ReseedMasterSST`), which fires only when
`Conf.Autorejoin` AND `Conf.Autoseed` are on. `autoseed` defaults to **false**, so a
reprov left the replicas empty — silent data loss.

## Change

`RollingReprov` now guarantees the reseed before destroying anything:
- **refuses to run if `autorejoin` is disabled** — `srv.go:947` gates the rejoin
  trigger on it, and forcing it globally would change failover behaviour for every
  other server, so we abort with a clear error rather than pilot it;
- **pilots `autoseed` on for the duration** and restores the operator's value after,
  so each reprovisioned replica is reseeded from the master.

Keeps the user doc's "rolling reprovision reseeds the replica from the primary" true.

## Test

- `go build ./cluster/...` green.
- Live validation on a dev OpenSVC cluster: see PR thread.

Closes #1771.


svaroqui:
Thanks for the review — all four points are well taken.

**Point 1 (concurrency / save-restore corruption) — fixed in `a562c13c2`.**
Added a dedicated `rollingReprovMutex` and `TryLock()` it at the very top of
`RollingReprov`, before any flag mutation. A concurrent run (double-fired API
POST, or the cron scheduler overlapping an operator run) is now refused with an
error rather than interleaved, so the `Autoseed`/`AutorejoinMysqldump`
save/restore can no longer capture another run's forced value or flip the
operator's setting permanently. `SetInRollingRestart` is only a status flag
(plain assignment, no check-and-set), so it could not serve as the guard.

**Point 3 (active-passive) — not applicable.** Rolling reprovision is not run on
active-passive topologies (the master is designated by topology handling, not
elected, and those clusters don't drive rolling ops), so the early
`ActivePassive` return in `RejoinMaster` isn't a path a reprov reaches. I'll
qualify the doc wording rather than add code.

**Point 2 (autoseed blast radius) — agreed, follow-up.** Piloting `autoseed`
cluster-wide for the (sequential, per-replica) duration does mean an unrelated
replica that crashes in that window would be auto-reseeded. The narrower fix is
to reseed only the specific servers being reprovisioned instead of flipping the
cluster-wide flag; tracking that as a follow-up rather than expanding this PR.

**Point 4 (regtest, T13) — agreed, follow-up.** A regression test asserting
(a) `RollingReprov` refuses when `autorejoin=false`, and (b) the piloted flags
are restored to their entry values, is the right T13 gate. Will add.

svaroqui:
**Point 4 (T13 regtest) addressed** in `03c1f3cfd` — and extended to all three rolling operations per request.

`regtest/test_rolling_reprov_reseed.go`: a shared `rollingHealthCheck` runs the op and asserts the cluster is left healthy — **same master**, **all slaves replicating** (an un-reseeded empty replica would not be), and **every server actually cycled** (uptime reset lower than before — guards against a silent no-op). Registered as `testRollingRestart` / `testRollingUpgrade` / `testRollingReprovReseed`.

Per functional-test discipline it gates on a **master-slave topology, not on the orchestrator** (the guarantee holds whatever the provisioning backend). Also adds `ServerMonitor.GetDatabaseUptime()` reading UPTIME from the collected status (dbhelper-normalised, no extra query).

Remaining follow-up: point 2 (narrow the autoseed pilot to only the reprovisioned servers).

svaroqui:
Superseded by #1773 — consolidated into a single rolling-improvement branch (same commits, plus the errorChan mutex).

PR #1784, opened 2026-09-09
fix(cluster): persist maintenance mode

### LINKED ISSUE #1783: Persist database-server maintenance mode across restart and config reload
## Problem

Database-server maintenance mode (`ServerMonitor.IsMaintenance`) is runtime-only. It
lives as an in-memory bool set by `SetMaintenance`/`DelMaintenance`/`SwitchMaintenance`
(`cluster/srv_set.go`, `cluster/srv_del.go`, `cluster/srv_tgl.go`) and is never
serialized. A `replication-manager` restart, or a live config reload, rebuilds every
`ServerMonitor` from scratch (`cluster/cluster.go` init + `ReloadConfig`, via
`cluster.newServerMonitor` in `cluster/srv.go`), which zero-values `IsMaintenance` back
to `false`. A server an operator intentionally pulled out of rotation silently rejoins
traffic after any restart — including maintenance set automatically by failover logic
when a slave can't safely rejoin (`cluster/cluster_fail.go`).

## Expected behavior

Maintenance state — whether set via the API/CLI/dashboard or by internal failover
logic — must survive process restart and config reload, and stay in effect until an
operator explicitly clears it.

## Proposed approach

- Add a durable per-cluster config field (`db-servers-maintenance-hosts`) alongside the
  existing `db-servers-ignored-hosts` membership pattern (`config.Config.IgnoreSrv`,
  matching helpers in `cluster/cluster_has.go`).
- `SetMaintenance`/`DelMaintenance`/`SwitchMaintenance` update both the runtime flag and
  the durable membership list, then request persistence via the existing
  `ConfigManager.SaveConfig` path (no new writer).
- `newServerMonitor` restores `IsMaintenance` from the configured membership at
  construction (both process startup and `ReloadConfig`), without replaying the
  maintenance state-change script.
- After proxies are rebuilt on startup/reload, replay `SetProxyServerMaintenance` for
  every restored maintenance server so HAProxy/ProxySQL/MaxScale converge immediately.

Full design notes: `doc/implementation/cluster/MAINTENANCE_PERSISTENCE.md` (added with
the implementing PR).

## Constraints

- Requires `monitoring-save-config=true` (the default) to be durable; documented as a
  prerequisite rather than silently no-op'd when disabled.
- No new API/CLI/GUI surface needed — existing maintenance endpoints already exist and
  route through the centralized lifecycle.

Branch: `persist-maintenance`.

caffeinated92:
## Summary

Fixes #1783 by persisting per-server maintenance mode across replication-manager configuration reloads and restarts.

Maintenance previously existed only on the in-memory `ServerMonitor.IsMaintenance` flag. Rebuilding server monitors during `InitFromConf()` reset that value, which could return a maintenance server to service after reload or restart.

### Durable configuration
- Adds the per-cluster `db-servers-maintenance-hosts` setting and `--db-servers-maintenance-hosts` flag.
- Persists maintenance membership through the existing dynamic cluster TOML save flow.

### Exact, lossless membership management
- Adds centralized maintenance token helpers shared by mutation and restoration paths.
- Uses exact URL/name matching rather than substring matching, preventing collisions such as `1.2.3.4:3306` matching `11.2.3.4:3306`.
- Edits the persisted membership list directly so entries for servers temporarily absent from the live cluster remain intact.

### Lifecycle and proxy reconciliation
- Updates set, clear, and toggle maintenance actions to synchronize durable membership with `IsMaintenance`.
- Applies the same durable behavior to operator-initiated and automatic failover-triggered maintenance.
- Restores maintenance directly when `newServerMonitor()` rebuilds monitors, without replaying database state-change scripts.
- Reconciles restored maintenance to rebuilt HAProxy, ProxySQL, and MaxScale backends immediately after `newProxyList()`.

### API and dashboard
- Documents durable maintenance behavior in the set, clear, and toggle API descriptions.
- Updates the dashboard confirmation copy so operators know maintenance stays active until explicitly cleared.

## Tests

- Adds focused cluster coverage for exact matching, substring-collision prevention, idempotent add/remove operations, mutation entry points, absent-server membership preservation, and rebuild restoration.
- Adds and registers `testMaintenancePersistReload`, which validates a real cluster reload sequence, persisted TOML membership, proxy backend convergence, clearing maintenance, and no state resurrection after a second reload.

## Validation

- `go test ./cluster/...`
- `go test ./config/...`
- `go test ./server/...`
- `go test ./regtest/...`
- Server build-tag combinations
- `git diff --check origin/develop...HEAD`

## Release gate

The new regtest covers the live reload/rebuild lifecycle. Execute the isolated Docker process-restart scenario before release to validate a fresh repman process reads the persisted cluster TOML and keeps the corresponding proxy backend in maintenance.

### DIFF
--- a/cluster/cluster.go
+++ b/cluster/cluster.go
@@ -858,6 +858,15 @@ func (cluster *Cluster) InitFromConf() {
 	if err != nil {
 		cluster.LogModulePrintf(cluster.Conf.Verbose, config.ConstLogModGeneral, config.LvlErr, "Could not set proxy list %s", err)
 	}
+	// Proxies were just rebuilt with no knowledge of maintenance restored onto
+	// Servers above (see newServerMonitor) -- replay the notification so
+	// HAProxy/ProxySQL/MaxScale converge immediately instead of waiting on
+	// their next refresh cycle.
+	for _, server := range cluster.Servers {
+		if server != nil && server.IsMaintenance {
+			cluster.SetProxyServerMaintenance(server.ServerID)
+		}
+	}
 	persistRebasedAppCreditCap := false
 	err = cluster.newAppList()
 	if err != nil {
--- a/cluster/cluster_get.go
+++ b/cluster/cluster_get.go
@@ -808,16 +808,6 @@ func (cluster *Cluster) GetIgnoredHostList() string {
 	return strings.Join(prevIgnored, ",")
 }
 
-// GetMaintenanceHostList returns the durable maintenance-host membership list
-// (cluster.Conf.MaintenanceSrv) directly. Unlike GetIgnoredHostList, this does
-// NOT reconstruct the list from currently-instantiated cluster.Servers: a host
-// temporarily absent from cluster.Servers (topology churn, a config-driven
-// host-list change) would otherwise have its persisted maintenance entry
-// silently dropped the next time membership is mutated.
-func (cluster *Cluster) GetMaintenanceHostList() string {
-	return cluster.Conf.MaintenanceSrv
-}
-
 func (cluster *Cluster) GetIgnoredROList() string {
 	var prevIgnoredRO []string
 	for _, server := range cluster.Servers {
--- a/cluster/cluster_get.go
+++ b/cluster/cluster_get.go
@@ -808,6 +808,16 @@ func (cluster *Cluster) GetIgnoredHostList() string {
 	return strings.Join(prevIgnored, ",")
 }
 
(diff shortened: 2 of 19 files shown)


caffeinated92:
@claude review my PR

caffeinated92:
@claude review my PR under 35 steps

caffeinated92:
@claude review my PR changes

caffeinated92:
@claude review my PR changes

PR #1788, opened 2026-09-11
fix: WriteBackupMetadata self-deadlock + JobsCheckStates api-mode 1146 storm

svaroqui:
## Context

Found while investigating **belair/db2** stuck in reseed for a day after an automatic failover during a prospect demo: the old master failed to auto-rejoin despite no binlog diff, a manual logical-backup rejoin was issued, and it never actually reseeded. pprof (`/debug/pprof/goroutine?debug=2`) showed goroutines blocked for ~a day on `server.backupMetaMutex`.

## Bug 1 — `WriteBackupMetadata` self-deadlock (`cluster/srv_job_backup.go`)

`WriteBackupMetadata` held `server.backupMetaMutex` across its **entire** body, including two unbounded poll loops:

```go
for task.State < 3 { time.Sleep(time.Second) }
...
for lastmeta.BinLogFileName == "" { time.Sleep(time.Second) }
```

`lastmeta.BinLogFileName` is set by the **writelog API path under the same mutex**. So the function blocked forever waiting on a value only settable by acquiring the lock it was already holding — a textbook self-deadlock. Every later backup-meta op then wedged behind it, including the flashback auto-rejoin's `snapshotLogicalBackupMeta` and the operator's manual logical rejoin. That is why db2 never reseeded.

**Fix:** release `backupMetaMutex` around each poll loop and re-acquire before mutating. All metadata mutations remain guarded; early returns precede the release, so the deferred `Unlock` stays balanced on every path.

## Bug 2 — `JobsCheckStates` missing api-mode guard (`cluster/srv_job.go`)

`JobsCheckStates` polled `replication_manager_schema.jobs` unconditionally. Under `scheduler-jobs-mode=api` that table doesn't exist, producing an `ERROR 1146` storm every tick. Its sibling `JobsCheckRunning` already guards this; `JobsCheckStates` now mirrors it and returns early when `SchedulerJobsMode == "api"`.

## Testing

- `go build ./cluster/ ./server/` clean.
- Also carries a docs-only commit: a debugging-discipline law (pprof how-to + never unblock before investigating fully) in `DEVELOPMENT_LAWS.md`.

Both are perpetual-monitoring blockers per the functional laws (a deadlock that halts backup-meta processing / a per-tick error storm).

### DIFF
--- a/cluster/backup_helpers_test.go
+++ b/cluster/backup_helpers_test.go
@@ -1089,3 +1089,51 @@ func TestMarkBackupPhysicalDoneRaceWithResticUpdate(t *testing.T) {
 		t.Fatalf("ResticSnapshotID = %q, want %q", server.LastBackupMeta.Physical.ResticSnapshotID, "snap-id")
 	}
 }
+
+// TestWaitForBinlogMetaNoDeadlock reproduces the WriteBackupMetadata self-deadlock: the
+// metadata writer held backupMetaMutex while polling for lastmeta.BinLogFileName, but that
+// field is only ever published by the writelog API path, which takes the SAME mutex. The
+// writer therefore blocked forever on a value it was preventing anyone from setting -- this
+// wedged belair/db2's rejoin for a day. waitForBinlogMeta must poll with the mutex RELEASED
+// so the publisher can make progress, then return holding it again.
+func TestWaitForBinlogMetaNoDeadlock(t *testing.T) {
+	_, server := newTestClusterServer(t)
+	lastmeta := &backupmgr.BackupMetadata{}
+
+	done := make(chan struct{})
+	go func() {
+		// Match WriteBackupMetadata's caller state: enter holding the mutex.
+		server.backupMetaMutex.Lock()
+		server.waitForBinlogMeta(lastmeta)
+		server.backupMetaMutex.Unlock() // helper returns holding it, per its contract
+		close(done)
+	}()
+
+	// The writelog API path publishes BinLogFileName under the same mutex. With the old code
+	// (poll while holding the lock) this Lock() would block forever -> deadlock.
+	go func() {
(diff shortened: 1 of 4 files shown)


svaroqui:
Thanks — all three points addressed in a50b92535:

1. **Data race on the polled reads.** Extracted the binlog wait into `waitForBinlogMeta` with a documented lock contract; it now reads `lastmeta.BinLogFileName` under a per-iteration `Lock`/`Unlock` (proper happens-before with the writelog writer, which sets it under the same mutex) rather than a bare read.
2. **Comment inaccuracy.** Corrected — the comment now states that `task.State` is advanced by `jobsUpdateState` and is **not** guarded by `backupMetaMutex` (a pre-existing unsynchronized read, not introduced here); only `BinLogFileName` is mutex-guarded, which is what the helper synchronizes around.
3. **Regression tests** (both green under `-race`, no warnings, alongside the existing `TestMarkBackupPhysicalDoneRaceWithResticUpdate`):
   - `TestWaitForBinlogMetaNoDeadlock` — reproduces the exact self-deadlock: caller holds the mutex, a second goroutine publishes `BinLogFileName` under the same mutex, asserts bounded-time return (hangs on the old code).
   - `TestJobsCheckStatesApiModeSkipsSQL` — api-mode + nil `Conn` must return `nil`, proving the guard short-circuits before the SQL/Conn path.

```
--- PASS: TestMarkBackupPhysicalDoneRaceWithResticUpdate (0.13s)
--- PASS: TestWaitForBinlogMetaNoDeadlock (1.00s)
--- PASS: TestJobsCheckStatesApiModeSkipsSQL (0.00s)
ok  github.com/signal18/replication-manager/cluster  7.828s
```

Left `task.State`'s synchronization as-is since it's pre-existing and out of scope for this deadlock fix.

PR #1791, opened 2026-09-14
gui(resource-manager): stable cluster order, per-layer empty series, legend labels

svaroqui:
## Problem

On dbaas-fr-2 the Resource Manager page showed:
- Consumed / Plan / Overcommit / Undercommit **DBU** graphs almost permanently on "No data"
- **APU** graphs flapping to "No data" every 30-60s
- raw graphite expressions (`removeBelowValue(diffSeries(...),0)`) printed as legend labels

Backend series and metric names were verified correct on production (`dbu.<cluster>.<host>.dbu`, `resourcemanager.<CLUSTER>.plan_dbu` exist and render in ~100ms). All causes were on the presentation path.

## Causes and fixes

1. **Unstable payload order.** `/api/global/resources` ranged over the `repman.Clusters` map, so the cluster order changed on every 10s poll. `ChartBarStack` keys its effect on the joined target list, re-ran it, aborted the in-flight fetch and drew the aborted (empty) result. → sort the payload by cluster name.
2. **First target used as time base.** `ChartBarStack` blanked the whole stack when the first target had no points. Only provisioned clusters have `dbu.*` / `plan_dbu` series (plan_dbu is emitted from the master), so any unprovisioned cluster landing first blanked the DBU graphs. → use the first target that has points; a target with no series (HTTP 200, empty body) is an empty layer, not a failure.
3. **Aborted poll drawn as empty.** → keep the previous data when a fetch is aborted.
4. **Legend derived from the path.** → `metricLabels` prop, the page passes cluster names.

## Verification

- `go build ./server/ ./cluster/` and real `npm run build` green.
- Deployed to s18/svc/repman (dbaas-fr-2) as v3.1.41-287-g37148e4c2: three consecutive `/api/global/resources` calls return an identical order; graphite proxy returns data for belair and an empty body for goodlands, chart renders belair with goodlands as an empty layer.

Side note found while investigating (not fixed here): `/api/global/resources` is registered only on the HTTP router (10001) and 404s on the HTTPS API port (10005), like the other `/api/global/*` routes.

https://claude.ai/code/session_01HyDkpec3it223qTSMpz31a


svaroqui:
Scope extended on the same branch (per Stéphane): the second commit fixes #1792 and the om3 cpu quota syntax.

- `WaitObjectConfigSettledV3`: after a config PUT, wait until the node's loaded instance config checksum matches the md5 of the file, before the rolling restart's start and before any pg update. Root cause and belair timeline in #1792.
- `OpenSVCCPUQuotaKeyword(cores)` = `"<cores*100>%@all"`; om3 rc20..rc36 divides by maxCpus unless `@all` is given, so this is the only form that means "N cores" on every node.
- `openSVCResizeCPU`: live CPU cap move on the PG slice when `prov-db-docker-run-args-limit` is off (v3).
- v3 template writes `pg_mem_limit` + `pg_cpu_quota` in DEFAULT when the docker cap is off.
- Unit test for the quota helper; doc section "om3 process-group facts" in RESOURCE_MANAGER.md.

Not yet validated live: needs a deploy on dbaas-fr-2 plus one rolling restart (settle wait) and a dev3 CPU resize (quota).

PR #1804, opened 2026-09-16
configurator: replication parallel workers are a setting (prov-db-replication-parallel-threads / -domain-parallel-threads), not the core count

svaroqui:
## Why

The configurator moduleset sets `slave_parallel_threads` and `slave_domain_parallel_threads` from `%%ENV:NODES_CPU_CORES%%`, so every provision / config re-render ties the replication workers to the core count. Parallel apply workers run on the thread pool and spend their time on IO and commit order: their number is the **concurrency needed to absorb network and commit latency**, not the cores (decision 2026-09-16; on dev3 a 2-core slave applied with 2 workers, and a per-domain cap of 2 silently limited a 32-worker pool to 2).

## What

Two new cluster settings (prov-db domain), exported to the templates:

| setting | default | template placeholder | variable |
|---|---|---|---|
| `prov-db-replication-parallel-threads` | 32 | `%%ENV:SVC_CONF_ENV_SLAVE_PARALLEL_THREADS%%` | `slave_parallel_threads` |
| `prov-db-replication-domain-parallel-threads` | 0 | `%%ENV:SVC_CONF_ENV_SLAVE_DOMAIN_PARALLEL_THREADS%%` | `slave_domain_parallel_threads` (0 = no per-domain cap, MariaDB default; a cap only for multi-domain topologies) |

Config fields, server flags, configurator get/set, cluster setters (`SetDBReplicationParallelThreads`, `SetDBReplicationDomainParallelThreads` — restart cookie like the other template-driven replication settings), API `settings/actions/set/...` cases, `GetEnv` entries, two gauges next to "Expire Binlog days" in Configs → DB.

## Companion step (not in this PR)

The compliance ruleset on the collector replaces `%%ENV:NODES_CPU_CORES%%` with the two placeholders above for `loose_slave_parallel_threads` / `loose_slave_domain_parallel_threads` in `replication.cnf` and in the start SQL; the moduleset export then lands in `share/opensvc/`. Until then the templates keep the core count and nothing changes.

`go build ./...`, `go test ./config/ ./cluster/configurator/`, `npm run build` green.

https://claude.ai/code/session_01SYNQWN5hjrXbn5iBGGz79E

PR #1805, opened 2026-09-16
monitor/gui: replication parallelism signal + Graphs page in sections with every Top gauge over time

svaroqui:
Tracks per server the average **binlog group commit size** over each tick (`Binlog_commits / Binlog_group_commits` deltas) — the commit concurrency a slave can apply in parallel (conservative-mode floor) — and the workers configured (`slave_parallel_threads` / `slave_parallel_workers`). Both exposed on the server JSON (`replicationGroupCommitSize`, `replicationParallelThreads`), emitted to Graphite as `mysql.<host>.replication_group_commit_size` / `replication_parallel_threads` (direct emission, not subject to the whitelist), and graphed on the cluster Graphs page as **Replication parallelism** (max across servers of each). This is the number to read before fixing the workers (PR #1804, decision: workers are concurrency, not cores) and the input of the future tuner (issue to follow).

`go build ./...` and `npm run build` green.
## Graphs page reorganised (folded in from #1807)
Five accordion sections instead of a flat wall of charts:
- **Workload**: Qps, Threads running / connected, Cpu thread pool, Cpu user stats, Network in/out, Schema size, Queries, Rows, Transactions
- **Replication**: Replication delay, Replication parallelism (above)
- **Resources**: Consumed DBU, Consumed APU
- **InnoDB**: Redo log, History list length, Read views, Mutex, Latch
- **Memory**: stacked memory chart, Swap, Cache miss

Every Top page gauge and per-instance bar graph now has a time series. The four "Cluster Workload" gauges without one are emitted per server as `mysql.<host>.workload_cpu_thread_pool_pct`, `workload_cpu_user_stats`, and from the master `workload_table_size_bytes` / `workload_index_size_bytes`. The five per-instance bar graphs (Queries, Rows, Transactions, Swap, Cache Miss) moved to `ServerMonitor.TopHeader()` and are emitted as `mysql.<host>.top_<graph>_<metric>`, so chart and Top page are the same computation (fixes the `COM_ISNERT` typo: the Top page Inserts bar was always 0). All first-class, not whitelist-gated (the per-cluster `whitelist.conf` is materialised once and never refreshed). `mysql_global_status_threads_connected` is added to `whitelist.conf.minimal` for the Connected line (existing clusters keep their materialised whitelist until `graphite-whitelist-template` is reapplied).

`GraphSection` is module-level on purpose: defined inside `Graphs` it would be a new component type every render and remount every chart.

Validated on dev3 (v3.1.41-303-g838f47dae): all 19 `top_*` series plus the workload ones reach Graphite.

https://claude.ai/code/session_01SYNQWN5hjrXbn5iBGGz79E


svaroqui:
Review follow-up (commit f903876fd):

1. **Stale value on a missed tick**: fixed. The computation is now `ServerMonitor.refreshReplicationParallelism`, which resets `ReplicationGroupCommitSize` to 0 before recomputing, so no previous snapshot / no group in the tick / counter reset all yield 0.
2. **Reuse `GetStatusDeltaValue`**: done, with the non-negative guard kept on top since that helper does not protect against a counter reset.
3. **MariaDB-only signal**: stated in the code comment, in the graph title ("commit concurrency, MariaDB only") and in the doc. MySQL/Percona show 0 for the group size; the workers line still comes from `slave_parallel_workers`.
4. **Unit tests**: `cluster/srv_replication_parallelism_test.go` covers normal delta, no group commit in the tick, counter reset, missed tick, first tick, MySQL workers.
5. **Docs (T8)**: `doc/implementation/cluster/GRAPHITE_SCOPE.md` gains a section on the first-class series (`replication_*`, `workload_*`, `top_*`) and the whitelist caveat. User doc on docs.signal18.io will follow with the graph captures.
6. **Issue (T9)**: this PR refs #1806 (the tuner this signal feeds).
7. **Regtest (T13)**: not applied here by the maintainer's decision: the change is graph wiring plus a read-only status delta with unit coverage; it was validated live on dev3 under sysbench (128 threads: group commit size max 1.21 at 23k qps).

PR #1810, opened 2026-09-17
resource sensor freshness gate: a reading older than 3 min withholds the dynamic resize (WARN0215, DBU series gap)

svaroqui:
## Problem
The consumed reading every dynamic-resize decision runs on is pushed by the dbjobs launcher once per run (~60 s), from the same script that then runs the dbjobs. During a long backup / reseed / optimize, or when the jobs container is down, the host keeps its **last** reading and nothing ages it: the instant axes keep the pre-job value, the `dbu.*` series repeats it every tick, and the sustained check sees a flat, fully present line. A pre-backup "under-used" reading can shrink the cap under the running backup (the quota is on the whole slice: the backup is throttled too); a "saturated" one can grow it every minute up to the overcommit ceiling on a load nobody measures. Only the memory apply was job-gated.

## Gate
- `DBUReading.ReceivedAt` stamped on ingest (repman clock, sensor clock skew irrelevant); `resourceSensorFreshnessWindow` = 3 min (three missed pushes), shared with the Kubernetes prerequisite check (was 5 min there).
- `CheckResourceConsumed`: a stale reading clears the four axes, so `DriveDynamicResize` and `driveDynamicShrink` stand still, and raises **WARN0215** with the age and likely cause. Re-set every tick while stale, resolves on the next push.
- `GetDatabaseMetrics`: no `dbu.*` series while stale. The gap is what the sustained coverage rule (≥ 80 % present samples) needs to withhold a decision that overlaps the silence. Never-measured and down servers keep their continuous series.
- A missing reading is not "stale": nil was already treated as unmeasured.

## Tests / docs
`TestResourceReadingStale`, `TestCheckResourceConsumedWithholdsOnStaleReading`, `TestGetDatabaseMetricsSkipsDBUSeriesWhenStale`. `RESOURCE_MANAGER.md` (decision loop, gate section, operator table). User doc on docs.signal18.io (Dynamic Resources: requirements, what you see, limits).

Follow-ups, not in this PR: push the reading from the launcher loop so a long job no longer starves it; gate both drivers on running jobs even with fresh readings.

https://claude.ai/code/session_01SYNQWN5hjrXbn5iBGGz79E


caffeinated92:
@claude review the PR again

caffeinated92:
@claude review the PR again

PR #1821, opened 2026-09-18
resize: the redo log follows a live memory step (MariaDB 10.9+, MySQL/Percona 8.0.30+), restart cookie otherwise

### LINKED ISSUE #1820: dynamic memory resize: resize the InnoDB redo log live with the buffer pool (MariaDB 10.9+ innodb_log_file_size, MySQL/Percona 8.0.30+ innodb_redo_log_capacity), restart cookie otherwise
`resizeMemorySQL` (cluster/cluster_resize_dynamic.go) is the live SET GLOBAL list of a memory step: buffer pool, key buffer, tmp/join buffers, max_session_mem_used. The redo log, the one configurator value of the memory model it leaves out (BP/4 as a power of two, floor 128 MB, cap 16 GB, #1749), stays at its boot-time size until a restart. After a shrink to 1 DBU the redo can again exceed the whole cgroup (the #1749 recovery problem); after a grow it stays small and throttles flushing. The template's `# mariadb_command:` line does not carry it either, so nothing ever resizes a running redo.

**Do:**
- MariaDB ≥ 10.9: `SET GLOBAL innodb_log_file_size = <MB>*1024*1024` (online redo resize, MDEV-27812), appended after the buffer pool on both grow and shrink (a shrink waits for the checkpoint and must not sit in front of the memory release).
- MySQL/Percona ≥ 8.0.30: `SET GLOBAL innodb_redo_log_capacity = <bytes>` (`innodb_log_file_size` is deprecated there).
- Older releases: the generated config already carries the new value, set the restart cookie and log it; never fail the memory step.
- Doc: a redo grow needs the new file's worth of free disk during the switch; a shrink under heavy writes can take a while (checkpoint) but does not block the workload.

Found while reviewing #1749 for v3.1.42: dev3 at 1 DBU runs redo 128 MB only because it restarted since 2026-09-02; an instance not restarted keeps the old 1024 MB redo.

### LINKED ISSUE #1822: dynamic memory resize: the in-flight gate compares the running buffer pool with the target the setter has just moved, so a real memory move is deferred as "still converging" before it starts
**Observed on dev3 (2026-09-18, v3.1.42-1-g546954eec, MariaDB 11.8.9, OpenSVC v3, `prov-db-dynamic-resource` on).** `POST /settings/actions/set/prov-db-memory/2048` from 768:

```
13:19:18 Live resource resize: memory grow (prov-db-dynamic-resource)
13:19:18 Live memory resize on db1.dev3.svc.cloud18:3306 deferred: a previous resize is still converging (in-flight gate)
13:19:18 Live memory resize on db2... deferred: a previous resize is still converging (in-flight gate)
13:19:18 Live memory resize on db3... deferred: a previous resize is still converging (in-flight gate)
resource resize applied=false dimension=memory direction=grow feasibility=yes statements="[]"   (x3)
```

Nothing was in flight: buffer pool 256 MB running and stable, no pending cgroup shrink. Setting back to 768 then ran the shrink SQL and the OpenSVC pg update normally, only because the running pool already equalled the new target.

**Cause.** `isMemoryResizeInFlight` (cluster/cluster_resize_dynamic.go, added by e29b618d7 on 2026-09-09 "in-flight gate -- never stack a memory resize on an unconverged one") returns true when the running `INNODB_BUFFER_POOL_SIZE` is more than 5 % away from `Configurator.GetConfigInnoDBBPSize()`. But `SetDBMemorySize` (and the automatic +1 DBU step, which goes through the same setter) moves the configurator target BEFORE `ResizeDynamicResources` runs. So at the moment of the gate the running pool is, by construction, far from the new target: every real memory move, grow or shrink, is deferred as "still converging", on every tick, forever. The only moves that pass are the ones where nothing has to change. Since 2026-09-09 the live memory resize has effectively been dead on every orchestrator (OpenSVC pg update, k8s in-place resize, script); CPU and IOPS are unaffected.

**Fix.** The gate must compare the running pool with the target that was actually ISSUED, not with the configurator's current wish: record the issued buffer-pool target on the server when the SET GLOBAL is sent (`applyConfirmedMemoryGrow`, the shrink phase 1) and clear it once the runtime is within 5 % of it; `isMemoryResizeInFlight` = pending cgroup shrink OR an issued target not yet reached. A target never issued is not in flight. Unit test: setter moves the target, gate must stay false; after an issued SET GLOBAL with runtime lagging, gate true; runtime converged, gate false.

Found while verifying #1820 (redo follows the memory step) on dev3; that PR's live check is blocked by this.

svaroqui:
Closes #1820. Closes #1822.

**1. The redo log follows a live memory step** (546954eec). `resizeMemorySQL` moved the buffer pool and the other memory-model values live but left the InnoDB redo log at its boot-time size until a restart. After a shrink to 1 DBU the redo could again exceed the whole memory cgroup (the #1749 recovery problem), after a grow it stayed small and throttled flushing.

- `redoLogResizeSQL`: `SET GLOBAL innodb_log_file_size = <MB>*1024*1024` on MariaDB ≥ 10.9 (online redo resize, MDEV-27812), `SET GLOBAL innodb_redo_log_capacity` on MySQL/Percona ≥ 8.0.30, valued from `GetConfigInnoDBLogFileSize` (BP/4 as a power of two, floor 128 MB, cap 16 GB). Appended **last**, after the buffer pool, on grow and on shrink: a redo shrink waits for the checkpoint and must not sit in front of the memory release.
- `redoLogRestartOnly`: a known older release cannot resize live, the generated config already carries the size, so the memory step sets the restart cookie. Unknown release: neither.
- Unit test per release; RESOURCE_MANAGER.md paragraph.

**2. The in-flight gate converges on the target actually issued** (8a09b2530, #1822, found by the dev3 check of 1). `isMemoryResizeInFlight` compared the running buffer pool with the configurator's latest wish, which the setter moves before the resize runs, so every real memory move was deferred as "still converging" on every tick. Now `ServerMonitor.IssuedBufferPoolBytes` records the target a resize actually sent; the gate is a pending cgroup shrink or an issued target the runtime has not reached within 5 %; converged clears it; nothing issued is not in flight. Unit test for the four states.

**3. Review follow-up** (679668638). `ExecScriptSQL` swallows every SQL error but 1238, so a redo statement failing silently would have left an oversized redo in a shrunk cgroup. `applyRedoLogResize` now runs the statement on its own right after the memory statements (same order), reads `innodb_log_file_size` / `innodb_redo_log_capacity` back, and sets the restart cookie on an exec error or a value not taken; restart-only releases get the cookie there. `resizeMemorySQL` no longer carries the statement. Test: the fallback (exec refused by sqlmock -> cookie) plus the per-release matrix.

**dev3 (MariaDB 11.8.9, OpenSVC v3, 1 DBU):**
- before the gate fix: 768 → 2048 deferred on all three nodes, "a previous resize is still converging", nothing in flight.
- after: 768 → 2048 applied live, buffer pool 256 → 512 MB, cgroup pg_mem_limit 2048 MB; 2048 → 768 back to 256 MB and 768 MB. `SET GLOBAL innodb_log_file_size = 128*1024*1024` issued after the buffer pool on every server, no error. - non-floor round trip (679668638, overcommit temporarily 600 % to pass the plan gate): 768 → 6144 MB, cgroup first, then buffer pool 256 MB → 2 GB and `SET GLOBAL innodb_log_file_size = 512*1024*1024`, runtime read back 536870912; 6144 → 768: buffer pool back to 256 MB, redo back to 134217728. No restart cookie, so both read-backs matched. The MySQL 8.0.30+ branch is unit-tested only.

Cautions for the doc: a redo grow needs the new file's worth of free disk during the switch; a shrink under heavy writes can take a while (checkpoint) but does not block the workload.


svaroqui:
Review follow-up addressed in 679668638: the redo statement now runs on its own after the memory statements, is read back, and any exec error or value not taken sets the restart cookie (test: exec refused -> cookie). Non-floor nit checked on dev3: 768 -> 6144 MB moved the redo 128 -> 512 MB live (read back 536870912) and back to 128 MB on the shrink.

